PR #2, opened 2019-11-07
Added a list of common pitfalls that for new developers

charles-lunarg:
In my experience, there are several common assumptions make about the Vulkan
API that aren't true. This is a short list of them, explaining how and why
one might reach that conclusion and why it isn't true or is unhelpful.

### DIFF
--- a/chapters/common_pitfalls.md
+++ b/chapters/common_pitfalls.md
@@ -1,10 +1,10 @@
 # Common Pitfalls for new Vulkan Developers
 
-This is a short list of commom assumptions and traps developers new to Vulkan can make.  
+This is a short list of common assumptions and traps developers new to Vulkan can make.  
 
 ### Validation Layers
 
-During developement, try to ensure that the Validation Layers are enabled. They are an invaluable tool for catching mistakes while using the Vulkan API. Parameter checking, object lifetimes, and threading violations all are part of the provided error checks. A way to reassure that they are enabled is to verify if the text "Debug Messenger Added" is in the output stream. More info can be found in the [Vulkan SDK](https://vulkan.lunarg.com/doc/sdk) documentation. It is first on the list because it is the first line of defense and should help reduce the time spent discovering where the mistake was made.
+During development, ensure that the Validation Layers are enabled. They are an invaluable tool for catching mistakes while using the Vulkan API. Parameter checking, object lifetimes, and threading violations all are part of the provided error checks. A way to reassure that they are enabled is to verify if the text "Debug Messenger Added" is in the output stream. More info can be found in the [Vulkan SDK](https://vulkan.lunarg.com/doc/sdk/latest/windows/layer_configuration.html) layer documentation. It is first on the list because it is the first line of defense and should help reduce the time spent discovering where the mistake was made.
 
 ### Vulkan is a box of tools
 
@@ -12,7 +12,7 @@ In Vulkan, there are often multiple solutions to a problem. Consider passing dat
 
 ### Recording command buffers  
 
-A common assumption when first learning about command buffers is that reusing them is paramount and that re-recording them every frame is costly. While it appear counterintuitive, there is often a greater cost in reuse than simply recording new command buffers each frame. The main cost is the additional complexity for managing dynamicism, frustum culling, and many other graphical effects. Many of the approaches used to reduce command buffer recording add non trivial state management or undue complexity to the rendering architecture. Therefore, especially in simpler situations, the cost of re-recording every frame is reasonable. Developers still concerned can and should profile command buffer recording to make better informed decisions about the best approach for their application.
+A common assumption when first learning about command buffers is that reusing them is paramount and that re-recording them every frame is costly. While it may appear counterintuitive, there is often a greater cost in reuse than simply recording new command buffers each frame. The main cost is the additional complexity for managing dynamic objects, frustum culling, and many other graphical effects. Many of the approaches used to reduce command buffer recording add non trivial state management or undue complexity to the rendering architecture. Therefore, especially in simpler situations, the cost of re-recording every frame is reasonable. Developers still concerned can and should profile command buffer recording to make better informed decisions about the best approach for their application.
 
 ### Multiple pipelines
 
@@ -24,7 +24,7 @@ A common technique to increase throughput is to make multiple instances of a res
 
 ### Multiple queues per queue family
 
-Several hardware platforms have more than one `VkQueue` per queue family. This can be useful by being able to submit work to the same queue family from seperate queues. While there can be advantages, it isn't necessarily better to create or use the extra queues. For performance recommendations, refer to hardware makers best practices guides.
+Several hardware platforms have more than one `VkQueue` per queue family. This can be useful by being able to submit work to the same queue family from separate queues. While there can be advantages, it isn't necessarily better to create or use the extra queues. For performance recommendations, refer to hardware makers best practices guides.
 
 ### Descriptor Sets
 
@@ -32,13 +32,13 @@ Descriptor Sets are designed to group data together by their usage and update fr
 
 ### Correct API usage practices
 
-While the Validation Layers can catch many types of errors, they are not perfect. Below is a short list of good habits to be in and possible sources of error when encountering odd behaviour.
+While the Validation Layers can catch many types of errors, they are not perfect. Below is a short list of good habits to be in and possible sources of error when encountering odd behavior.
 
 * Initialize all variables and structs.
 * Use the correct `sType` for each structure.
 * Verify correct `pNext` chain usage, nulling it out when not needed.
 * There are no default values in Vulkan.
 * Use correct enum, `VkFlag`, and bitmask values. 
-* Considering using a type-safe Vulkan wrapper, eg [Vulkan.hpp](https://github.com/KhronosGroup/Vulkan-Hpp) for C++
+* Consider using a type-safe Vulkan wrapper, eg [Vulkan.hpp](https://github.com/KhronosGroup/Vulkan-Hpp) for C++
 * Check function return values, eg `VkResult`.
 * Call clean up functions where appropriate.
--- a/chapters/common_pitfalls.md
+++ b/chapters/common_pitfalls.md
@@ -1,22 +1,22 @@
 # Common Pitfalls for new Vulkan Developers
 
-This is a short list of common assumptions and traps developers new to Vulkan can make.  
+This is a short list of common assumptions and traps developers new to Vulkan can make. 
 
 ### Validation Layers
 
-During development, ensure that the Validation Layers are enabled. They are an invaluable tool for catching mistakes while using the Vulkan API. Parameter checking, object lifetimes, and threading violations all are part of the provided error checks. A way to reassure that they are enabled is to verify if the text "Debug Messenger Added" is in the output stream. More info can be found in the [Vulkan SDK](https://vulkan.lunarg.com/doc/sdk/latest/windows/layer_configuration.html) layer documentation. It is first on the list because it is the first line of defense and should help reduce the time spent discovering where the mistake was made.
+During development, ensure that the Validation Layers are enabled. They are an invaluable tool for catching mistakes while using the Vulkan API. Parameter checking, object lifetimes, and threading violations all are part of the provided error checks. A way to reassure that they are enabled is to verify if the text "Debug Messenger Added" is in the output stream. More info can be found in the [Vulkan SDK](https://vulkan.lunarg.com/doc/sdk/latest/windows/layer_configuration.html) layer documentation.
 
 ### Vulkan is a box of tools
 
-In Vulkan, there are often multiple solutions to a problem. Consider passing data to a shader, one can use uniform buffers, storage buffers, vertex attributes, instance buffer, push constants, and more. Are all valid methods each with their own benefits and drawbacks. There is rarely a "perfect" solution to any problem. Therefore, use the tools available and create a good solution instead of trying to find a perfect solution.
+In Vulkan, most problems can be tackled with multiple methods, each with their own benefits and drawbacks. There is rarely a "perfect" solution and obsessing over finding one is often a fruitless effort. When faced with a decision, make an informed choice and create an adequate solution that meets the current needs. To become informed, use this guide, reference a hardware vendors best practices guide, and profile various solutions.  
 
 ### Recording command buffers  
 
-A common assumption when first learning about command buffers is that reusing them is paramount and that re-recording them every frame is costly. While it may appear counterintuitive, there is often a greater cost in reuse than simply recording new command buffers each frame. The main cost is the additional complexity for managing dynamic objects, frustum culling, and many other graphical effects. Many of the approaches used to reduce command buffer recording add non trivial state management or undue complexity to the rendering architecture. Therefore, especially in simpler situations, the cost of re-recording every frame is reasonable. Developers still concerned can and should profile command buffer recording to make better informed decisions about the best approach for their application.
+Many early Vulkan tutorials and documents recommended writing a command buffer once and re-using it wherever possible. However, in practice re-use rarely has a performance benefit and incurs a non-trivial development burden. While it may appear counterintuitive, as re-using computed data is a common optimization, managing a scene with objects being added and removed as well as techniques such as frustum culling which vary draw calls on a per frame basis make reusing command buffers a serious design challenge. Instead prefer to simply re-record fresh command buffers every frame.  
 
 ### Multiple pipelines
 
-A graphics `VkPipeline` contains the combination of state needed to perform a draw call. For every input combination (shaders, vertex layout, primitive assembly, depth testing, blending mode, etc), a new pipeline is needed. This necessitates creating and binding many pipelines for complex rendering situations. While it might appear beneficial to try to create as few pipeline as possible by increasing complexity elsewhere, it is best to profile and determine if this is truly a big performance penalty in the first place.
+A graphics `VkPipeline` contains the combination of state needed to perform a draw call. For every input combination (shaders, vertex layout, primitive assembly, depth testing, blending mode, etc), a new pipeline is needed. This necessitates creating and binding many pipelines for complex rendering situations. While it might appear beneficial to try to create as few pipelines as possible by increasing complexity elsewhere, it is best to profile and determine if this is truly a big performance penalty in the first place.
 
 ### Resource duplication per swapchain image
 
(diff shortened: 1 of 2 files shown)


chapters/common_pitfalls.md
```diff
@@ -0,0 +1,43 @@
+# Common Pitfalls for new Vulkan Developers
+
+This is a short list of commom assumptions and traps developers new to Vulkan can make.  
+
+### Validation Layers
+
+During developement, try to ensure that the Validation Layers are enabled. They are an invaluable tool for catching mistakes while using the Vulkan API. Parameter checking, object lifetimes, and threading violations all are part of the provided error checks. A way to reassure that they are enabled is to verify if the text "Debug Messenger Added" is in the output stream. More info can be found in the Vulkan SDK documentation. It is first in the list because it is the first line of defense and should help reduce the time spent discovering where the mistake was made.
```
sfricke-samsung: > in the Vulkan SDK documentation.

Could you please provide a link to there

sfricke-samsung: > It is first in the list

Can change that to

> It is first on this list

chapters/common_pitfalls.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Recording command buffers  
+
+A common assumption when first learning about command buffers is that reusing them is paramount and that re-recording them every frame is costly.  While it appear counterintuitive, there is often a greater cost in reuse than simply recording new command buffers each frame. The main cost is the additional complexity for managing dynamicism, frustum culling, and many other graphical effects. Many of the approaches used to reduce command buffer recording add non trivial state management or undue complexity to the rendering architecture. Therefore, especially in simpler situations, the cost of re-recording every frame is reasonable. Developers still concerned can and should profile command buffer recording to make better informed decisions about the best approach for their application.
```
sfricke-samsung: super nit: single space please
> costly.  While

chapters/common_pitfalls.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Multiple pipelines
+
+A `VkPipeline` contains the combination of state needed to perform a draw call. For every input combination (shaders, vertex layout, primitive assembly, depth testing, blending mode, etc), a new pipeline is needed. This necessitates creating and binding many pipelines for complex rendering situations. While it might appear beneficial to try to create as few pipeline as possible by increasing complexity elsewhere, it is best to profile and determine if this is truly a big performance penalty in the first place.
```
sfricke-samsung: >A `VkPipeline`
to
>A graphics `VkPipeline`

chapters/common_pitfalls.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Multiple queues per queue family
+
+Several hardware platforms have more than one `VkQueue` per queue family. This can be useful by being able to submit work to the same queue family from seperate queues. While there can be advantages, it isn't necessarily better to create or use the extra queues. For performance reccomendations, refer to hardware makers best practices guides.
```
sfricke-samsung: reccomendations -> recommendations

chapters/common_pitfalls.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Descriptor Sets
+
+Descriptor Sets are designed to group data together by their usage and update frequency. It is reasonable to have a descriptor set bound for each of group of resources needed in a shader, for example object matrices, lights, textures, material information, and per-instance data. Vulkan requires hardware to be able to support at least 4 concurrently bound descriptor sets, as described by the maxBoundDescriptorSets value in `VkPhysicalDeviceLimits`, with many supporting 8 or more. Therefore is safe to use more than 1 descriptor set at a time. While there may be situations that may make a single descriptor set advantageous, refer to a best practices guide for specific details.
```
sfricke-samsung: > each of group of resources

to

> each group of resources

chapters/common_pitfalls.md
```diff
@@ -0,0 +1,43 @@
...
+### Resource duplication per swapchain image
+
+A common technique to increase throughput is to make multiple instances of a resource, where each instance can be used while rendering a different frame. The swapchain is one such example, as it contains multiple images. This can lead to the assumption that the number of instances each resource has should be the same as the swapchain image count. The issue with this assumption is that very few resources need to be duplicated the same amount as the swapchain image count. While it useful for some things, for example the number of semaphores used to signal when a swapchain image is available, it is worth understanding which resources have one to one relationships and which don't.
+
+### Vulkan is a box of tools
+
+In Vulkan, there are often multiple solutions to a problem. Consider passing data to a shader. One can use uniform buffers, storage buffers, vertex attributes, instance buffer, push constants, and more. Are all valid methods each with their own benefits and drawbacks. There is rarely a "perfect" solution to any problem. Therefore, use the tools available and create a good solution instead of trying to find a perfect solution.
```
sfricke-samsung: Consider passing data to a shader, one

sfricke-samsung: These are all valid methods...

chapters/common_pitfalls.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Correct API usage practices
+
+While the Validation Layers can catch many types of errors, they are not perfect. Below is a short list of good habits to be in and possible sources of error when encountering odd behaviour.
+
+* Initialize all variables and structs.
```
sfricke-samsung: This brings up a possible other point that we should express in this section that there are no defaults in Vulkan for structs

chapters/common_pitfalls.md
```diff
@@ -0,0 +1,43 @@
...
+* Use the correct `sType` for each structure.
+* Verify correct `pNext` chain usage, nulling it out when not needed.
+* Use correct enum, `VkFlag`, and bitmask values. 
+* Considering using a type-safe Vulkan wrapper, eg Vulkan.hpp for C++
```
sfricke-samsung: link to Vulkan.hpp please

chapters/common_pitfalls.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Resource duplication per swapchain image
+
+A common technique to increase throughput is to make multiple instances of a resource, where each instance can be used while rendering a different frame. The swapchain is one such example, as it contains multiple images. This can lead to the assumption that the number of instances each resource has should be the same as the swapchain image count. The issue with this assumption is that very few resources need to be duplicated the same amount as the swapchain image count. While it useful for some things, for example the number of semaphores used to signal when a swapchain image is available, it is worth understanding which resources have one to one relationships and which don't.
```
sfricke-samsung: > While it useful

to

> While it is useful

chapters/common_pitfalls.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Vulkan is a box of tools
```
sfricke-samsung: Could we make this item number 2 on the list, I feel its that important

chapters/common_pitfalls.md
```diff
@@ -0,0 +1,44 @@
...
+In Vulkan, there are often multiple solutions to a problem. Consider passing data to a shader, one can use uniform buffers, storage buffers, vertex attributes, instance buffer, push constants, and more. Are all valid methods each with their own benefits and drawbacks. There is rarely a "perfect" solution to any problem. Therefore, use the tools available and create a good solution instead of trying to find a perfect solution.
+
+### Correct API usage practices
+
+While the Validation Layers can catch many types of errors, they are not perfect. Below is a short list of good habits to be in and possible sources of error when encountering odd behaviour.
+
+* Initialize all variables and structs.
+* Use the correct `sType` for each structure.
+* Verify correct `pNext` chain usage, nulling it out when not needed.
+* There are no default values in vulkan.
```
sfricke-samsung: nit: `Vulkan`

chapters/common_pitfalls.md
charles-lunarg: I added the line "There are no default values in vulkan."
I tried to keep the items in the list short, but this needs more info. Unfortunately I couldn't quite decide what to add to the line.

chapters/common_pitfalls.md
charles-lunarg: I feel this section needs rewriting. It lacks a clear description of what is being stated. 
My goal is to describe that Vulkan offers a lot of tools, that they are sometimes suited for a particular job, but there isn't always a specific tool for every job and developers will need to create a solution given the tools rather than rely on the API to provide the exact solution.

sfricke-samsung: I really liked the `There is rarely a "perfect" solution to any problem.` statement and why I wanted this to be 2nd on the list as its a good overall "warning"

chapters/common_pitfalls.md
charles-lunarg: Ah, I misunderstood which was second.
Will fix that in the next commit.

sfricke-samsung:
@charles-lunarg I think this all looks good to me, but want someone else at an IHV to review this as well

chapters/common_pitfalls.md
```diff
@@ -0,0 +1,44 @@
+# Common Pitfalls for new Vulkan Developers
+
+This is a short list of commom assumptions and traps developers new to Vulkan can make.  
+
+### Validation Layers
+
+During developement, try to ensure that the Validation Layers are enabled. They are an invaluable tool for catching mistakes while using the Vulkan API. Parameter checking, object lifetimes, and threading violations all are part of the provided error checks. A way to reassure that they are enabled is to verify if the text "Debug Messenger Added" is in the output stream. More info can be found in the [Vulkan SDK](https://vulkan.lunarg.com/doc/sdk) documentation. It is first on the list because it is the first line of defense and should help reduce the time spent discovering where the mistake was made.
```
DethRaid: "try to ensure" -> "ensure". I'd also include a link to a tutorial on how to enable the validation layers programmatically

chapters/common_pitfalls.md
```diff
@@ -0,0 +1,44 @@
...
+
+### Recording command buffers  
+
+A common assumption when first learning about command buffers is that reusing them is paramount and that re-recording them every frame is costly. While it appear counterintuitive, there is often a greater cost in reuse than simply recording new command buffers each frame. The main cost is the additional complexity for managing dynamicism, frustum culling, and many other graphical effects. Many of the approaches used to reduce command buffer recording add non trivial state management or undue complexity to the rendering architecture. Therefore, especially in simpler situations, the cost of re-recording every frame is reasonable. Developers still concerned can and should profile command buffer recording to make better informed decisions about the best approach for their application.
```
DethRaid: "While it appear" -> "while it may appear"

chapters/common_pitfalls.md
```diff
@@ -0,0 +1,44 @@
...
+
+### Multiple pipelines
+
+A graphics `VkPipeline` contains the combination of state needed to perform a draw call. For every input combination (shaders, vertex layout, primitive assembly, depth testing, blending mode, etc), a new pipeline is needed. This necessitates creating and binding many pipelines for complex rendering situations. While it might appear beneficial to try to create as few pipeline as possible by increasing complexity elsewhere, it is best to profile and determine if this is truly a big performance penalty in the first place.
```
martty: as few pipeline*s*

chapters/common_pitfalls.md
```diff
@@ -0,0 +1,44 @@
...
+While the Validation Layers can catch many types of errors, they are not perfect. Below is a short list of good habits to be in and possible sources of error when encountering odd behavior.
+
+* Initialize all variables and structs.
+* Use the correct `sType` for each structure.
+* Verify correct `pNext` chain usage, nulling it out when not needed.
+* There are no default values in Vulkan.
+* Use correct enum, `VkFlag`, and bitmask values. 
+* Consider using a type-safe Vulkan wrapper, eg [Vulkan.hpp](https://github.com/KhronosGroup/Vulkan-Hpp) for C++
+* Check function return values, eg `VkResult`.
+* Call clean up functions where appropriate.
```
martty: cleanup

chapters/common_pitfalls.md
```diff
@@ -0,0 +1,44 @@
...
+
+### Validation Layers
+
+During development, ensure that the Validation Layers are enabled. They are an invaluable tool for catching mistakes while using the Vulkan API. Parameter checking, object lifetimes, and threading violations all are part of the provided error checks. A way to reassure that they are enabled is to verify if the text "Debug Messenger Added" is in the output stream. More info can be found in the [Vulkan SDK](https://vulkan.lunarg.com/doc/sdk/latest/windows/layer_configuration.html) layer documentation. It is first on the list because it is the first line of defense and should help reduce the time spent discovering where the mistake was made.
+
+### Vulkan is a box of tools
+
+In Vulkan, there are often multiple solutions to a problem. Consider passing data to a shader, one can use uniform buffers, storage buffers, vertex attributes, instance buffer, push constants, and more. Are all valid methods each with their own benefits and drawbacks. There is rarely a "perfect" solution to any problem. Therefore, use the tools available and create a good solution instead of trying to find a perfect solution.
```
Tobski: This paragraph is a little weird without at least some pointer as to how to judge the best tool for the job. I think some indication of where to find resources to help (e.g. this guide!) is a good idea.

chapters/common_pitfalls.md
```diff
@@ -0,0 +1,44 @@
...
+
+### Recording command buffers  
+
+A common assumption when first learning about command buffers is that reusing them is paramount and that re-recording them every frame is costly. While it may appear counterintuitive, there is often a greater cost in reuse than simply recording new command buffers each frame. The main cost is the additional complexity for managing dynamic objects, frustum culling, and many other graphical effects. Many of the approaches used to reduce command buffer recording add non trivial state management or undue complexity to the rendering architecture. Therefore, especially in simpler situations, the cost of re-recording every frame is reasonable. Developers still concerned can and should profile command buffer recording to make better informed decisions about the best approach for their application.
```
Tobski: I think it would be better to pretty much rule out command buffer reuse altogether - more detailed chapters could explain the few cases where they are actually useful. Their utility is pretty low in practice - I think the main real world use cases are things like pre-canned layout transitions to stitch command buffers together, or dedicated post-processes.

I'd maybe reword this to start with something like:

> Many early Vulkan tutorials and documents recommend command buffer reuse - but outside of toy examples, this is largely unnecessary and puts significant additional burden on the developer. The gains you might see from pre-recorded command buffers is typically vanishingly small due to the already low overhead of Vulkan. The main cost is the additional complexity for managing...

chapters/common_pitfalls.md
Tobski: I'm unclear what this section is trying to advocate. What might make more sense here is to say "rely on pipeline caches to reduce duplication, rather than managing duplication yourself." I think this section needs a big old caveat though that "for large engines, it may become necessary to take a more complex approach - but when starting out this is unlikely to be a bottleneck."

chapters/common_pitfalls.md
```diff
@@ -0,0 +1,44 @@
...
+
+### Resource duplication per swapchain image
+
+A common technique to increase throughput is to make multiple instances of a resource, where each instance can be used while rendering a different frame. The swapchain is one such example, as it contains multiple images. This can lead to the assumption that the number of instances each resource has should be the same as the swapchain image count. The issue with this assumption is that very few resources need to be duplicated the same amount as the swapchain image count. While it is useful for some things, for example the number of semaphores used to signal when a swapchain image is available, it is worth understanding which resources have one to one relationships and which don't.
```
Tobski: I think this needs a sentence or two explaining that you only need one of a given resource if all uses are temporally separated - e.g. if a depth buffer is used in one frame, then the data is never used again after that frame, then the next frame can use the same depth buffer.

chapters/common_pitfalls.md
```diff
@@ -0,0 +1,44 @@
...
+
+### Multiple queues per queue family
+
+Several hardware platforms have more than one `VkQueue` per queue family. This can be useful by being able to submit work to the same queue family from separate queues. While there can be advantages, it isn't necessarily better to create or use the extra queues. For performance recommendations, refer to hardware makers best practices guides.
+
+### Descriptor Sets
+
+Descriptor Sets are designed to group data together by their usage and update frequency. It is reasonable to have a descriptor set bound for each group of resources needed in a shader, for example object matrices, lights, textures, material information, and per-instance data. Vulkan requires hardware to be able to support at least 4 concurrently bound descriptor sets, as described by the maxBoundDescriptorSets value in `VkPhysicalDeviceLimits`, with many supporting 8 or more. Therefore is safe to use more than 1 descriptor set at a time. While there may be situations that may make a single descriptor set advantageous, refer to a best practices guide for specific details.
```
Tobski: nit: "Therefor *it* is safe..."

Also general comment on this paragraph; it kind of feels like you're just trying to pad the words out. The salient point here (that using multiple descriptor sets is fine) could probably be done in two sentences, and it probably should be.

chapters/common_pitfalls.md
```diff
@@ -0,0 +1,44 @@
+# Common Pitfalls for new Vulkan Developers
+
+This is a short list of common assumptions and traps developers new to Vulkan can make.
```
Tobski: One thing that would be good to add here is a piece of general advice that you don't need to aim for 100% perfection - you have to judge what's good enough for your use case. Code in the way that feels natural, follow this guide, watch out for pitfalls, and only if you start seeing poor performance you should use perf recommendations and PGO to figure out how to improve - don't twist yourself in knots to get things working a particular way unless you need to.

Tobski: (Noting that this is a common theme for this whole section)

charles-lunarg:
@sfricke-samsung After all the comments are addressed, should the commits be squashed down into 1, to better manage git history?

sfricke-samsung:
@charles-lunarg yes, but I know GitHub has a "Squash and Merge" option so happy to do that myself (I was planning to squash it regardless unless there was a reason to conserve the commit). You are more then welcome to squash yourself too

charles-lunarg:
Ah I will leave it to you. No reason to manually do the work when a tools that does it automatically exists.

chapters/common_pitfalls.md
AttilioProvenzano-ARM: The idea of "temporally separated" is trickier with pipelined GPUs, fragment work may run in parallel with vertex work for the following frame. The depth buffer example still holds true because it only involves fragment and depth tests, but if reusing a resource creates a FRAG -> VERT dependency, that would create bubbles in the pipeline.

charles-lunarg: This is the one point that is rather hard to write because there is a lot of nuance and jargon at play. 
My main point succinctly put is "Few resources need to be duplicated the same amount as the swapchain image count". IE, I have X swapchain images, therefore I should have X everything else. 
There are so many ways to solve the topic of resource contention and pipelining frames that it deserves its own page. As thus I am just trying to highlight that not everything requires 1:1.

chapters/common_pitfalls.md
charles-lunarg: Reworked this but not sure if it addressed the commented issues.

chapters/common_pitfalls.md
charles-lunarg: I opted to put many of the points you made into the "Vulkan is a tool box" section, since it had a lot of overlap over how to create a well rounded solution.

charles-lunarg:
I have re-written large sections of this article to better address the mentioned issues. Hopefully it better explains the intended point.

sfricke-samsung:
@charles-lunarg I really like the rewrite and I think it looks good. Will wait on @Tobski 👍 and then let's get this merged in!

chapters/common_pitfalls.md
```diff
@@ -0,0 +1,44 @@
...
+
+### Multiple queues per queue family
+
+Several hardware platforms have more than one `VkQueue` per queue family. This can be useful by being able to submit work to the same queue family from separate queues. While there can be advantages, it isn't necessarily better to create or use the extra queues. For specific performance recommendations, refer to hardware makers best practices guides.
```
sfricke-samsung: super nit: "hardware vendors' best practice guides"

PR #11, opened 2019-11-14
Update tools entry

PatrickPurcell:
This PR adds an entry for Intel(R) GPA to `development_tools.md`


sfricke-samsung:
thanks for addition

PR #21, opened 2020-02-03
fix typo

mickbeaver:
The online man pages are now for the 1.2 spec.


sfricke-samsung:
great catch @mickbeaver , thanks for making the PR!

PR #22, opened 2020-02-04
Fixed typo.

mickbeaver:
* will only exists -> will only exist


sfricke-samsung:
Thanks again @mickbeaver

PR #40, opened 2020-05-08
Add Mapping Data to Shaders chapter

sfricke-samsung:


chapters/mapping_data_to_shaders.md
```diff
@@ -0,0 +1,308 @@
...
+
+> Note: All SPIR-V assembly was generated with glslangValidator
+
+This chapter goes over how to [interface Vulkan with SPIR-V](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#interfaces) in order to map data. Using the `VkDeviceMemory` objects allocated from `vkAllocateMemory`, it is up to the application to properly map the data from Vulkan in a way that the SPIR-V shader understands how to consume it correctly.
+
+In core Vulkan, there are 4 fundamental ways to map data from your Vulkan application to interface with SPIR-V:
+  - [Input Attributes](#input-attributes)
+  - [Descriptors](#descriptors)
+  - [Push Constants](#push-constants)
+  - [Specialization Constants](#specialization-constants)
```
Tobski, line 11: I think you need to mention buffer pointers. They’re a powerful tool for a range of use cases, and not mentioning them feels wrong to me.

chapters/mapping_data_to_shaders.md
```diff
@@ -0,0 +1,308 @@
...
+
+The following results would look as followed
+
+![mapping_data_to_shaders_descriptor_2.png](../images/mapping_data_to_shaders_descriptor_2.png)
+
+
+## Push Constants
+
+A push constant is a small bank of values accessible in shaders. Push constants allow the application to set values used in shaders without creating buffers or modifying and binding descriptor sets for each update.
+
```
Tobski, line 142: It would be good to explain what you might use them for. Push constants in particular should be recommended for a small amount (a few dwords) of high frequency data (e.g. indices).

chapters/mapping_data_to_shaders.md
```diff
@@ -0,0 +1,308 @@
...
+            myResults = false;
+        }
+        // ...
+    }
+    ```
+2. Improving backend optimizations
+    - The "backend" here refers the the implementation's compiler that takes the resulting SPIR-V and lowers it down to some ISA to run on the device.
+    - Constant values allow a set of optimizations such as [constant folding](https://en.wikipedia.org/wiki/Constant_folding), [dead code elimination](https://en.wikipedia.org/wiki/Dead_code_elimination), etc. to occur.
+    - While divergence gives poor performance in shader code, if the compiler knows from the SPIR-V specialization constant that only one branch is possible, it can easily eliminate the branch for you.
+3. Effecting types and memory sizes
```
Tobski, line 275: Nit: Should be “Affecting“

chapters/mapping_data_to_shaders.md
```diff
@@ -0,0 +1,308 @@
...
+        } else {
+            myResults = false;
+        }
+        // ...
+    }
+    ```
+2. Improving backend optimizations
+    - The "backend" here refers the the implementation's compiler that takes the resulting SPIR-V and lowers it down to some ISA to run on the device.
+    - Constant values allow a set of optimizations such as [constant folding](https://en.wikipedia.org/wiki/Constant_folding), [dead code elimination](https://en.wikipedia.org/wiki/Dead_code_elimination), etc. to occur.
+    - While divergence gives poor performance in shader code, if the compiler knows from the SPIR-V specialization constant that only one branch is possible, it can easily eliminate the branch for you.
```
Tobski, line 274: This isn’t strictly true, and is something of an artefact of days past. The cases where specialisation constants could otherwise be used would result in uniform branching - which is extremely low cost. Only non-uniform branches are actively expensive, and you can’t replace those with specialisation constants.

In cases where specialising would result in a high number of pipeline objects, it’s often a better idea to perform dynamic uniform branching to avoid the compilation cost.

PR #81, opened 2021-02-15
Remove "Adjusting format" section

abbaswasim:
The adjusting format section is wrong. UV will point at pad instead. Its easier to just remove this section.

### DIFF
--- a/chapters/vertex_input_data_processing.md
+++ b/chapters/vertex_input_data_processing.md
@@ -91,14 +91,7 @@ struct Vertex {
 };
 ```
 
-There are 2 ways to go about this by either:
-
-- Adjusting the location **offset**
-- Adjusting the location **format**
-
-#### Adjusting offset
-
-For the first case of changing the offset, the only change to your pipeline creation becomes
+In this case you would need to adjust the offset to point at the right attribute. The only change to your pipeline creation becomes
 
 ```patch
         1,                          // location
@@ -110,23 +103,9 @@ For the first case of changing the offset, the only change to your pipeline crea
 
 ![vertex_input_data_processing_example_b_offset](../images/vertex_input_data_processing_example_b_offset.png)
 
-#### Adjusting format
-
-For the second case changing the format, you keep the `3 * sizeof(float)` offset and adjust the format to change the size of each location element
-
-```patch
-        0,                          // location
-        binding.binding,            // binding
--        VK_FORMAT_R32G32B32_SFLOAT, // format
-+        VK_FORMAT_R32G32B32A32_SFLOAT, // format
-        0                           // offset
-```
-
 Note, this does **not** require an update to the shader from a `vec3` to a `vec4` as the spec states
 
 > If the vertex shader has fewer components, the extra components are discarded.
-
-![vertex_input_data_processing_example_b_format](../images/vertex_input_data_processing_example_b_format.png)
 
 ### Example C - non-interleaved
 


sfricke-samsung:
I am confused by what you mean by "UV will point at pad instead"

The spec says

> If the vertex shader has fewer components, the extra components are discarded.

so if the format is moved to `VK_FORMAT_R32G32B32A32_SFLOAT` I am not sure why UV is wrong

abbaswasim:
The spec describes `offset` in `VkVertexInputAttributeDescription` as:

> `offset` is a byte offset of this attribute relative to the start of an element in the vertex input binding.

So IIUC no matter what your format is, the offset has to be correct to point at the correct byte offset in your struct. So for the following Vertex.

```c
struct Vertex {
    float   x, y, z, pad;
    uint8_t u, v;
 };
 ```

using an offset of `3 * sizeof(float)` will point at `pad` not `u` from `uint8_t u, v;` which is wrong. Imagine your struct is like the following: [silly example but hopefully gets the point across]

```c
struct Vertex {
    float   x, y, z, pad[125];
    uint8_t u, v;
 };
 ```
In this case you won't find a `VkFormat` for your first attribute  that you can specify which will take you to the `u` offset.

> If the vertex shader has fewer components, the extra components are discarded.

True. But not the issue here. The issue is that the offset is wrong. My worry is not that 

```glsl
layout(location = 0) in vec3 inPos;
```

is getting 4 float values or not but rather.

```glsl
layout(location = 1) in uvec2 inUV;
```
is not getting the correct values.

Hope that makes sense.

sfricke-samsung:
So using your example with `VK_FORMAT_R32G32B32A32_SFLOAT`, then the developer just needs to update where in the `Vertex` stride to grab the second attribute

```cpp
attributes[1].offset = 124 * sizeof(float);
```

From spec:

> For each attribute, raw data is extracted starting at attribAddress and is converted from the VkVertexInputAttributeDescription’s format

The `format` is really there to say

"This Input Attribute is the size of this format and that much raw data will be read from the vertex buffer"

> In this case you won't find a VkFormat for your first attribute that you can specify which will take you to the u offset.

So the only reason I have that example in was to demonstrate **what** happens when the format is larger then the shader value. This came from a real world example of confusion of the needle-in-the-haystack spec text that described what happens if `VK_FORMAT_R32G32B32A32_SFLOAT` is lined up with a `vec3` attribute

sfricke-samsung:
To add to this, I am pretty sure it also legal to go

```cpp
struct Vertex {
    float   x, y, z;
};
```
with
```glsl
layout(location = 0) in vec3 inA;
layout(location = 1) in vec2 inB;
```
have the offset both to 0
```cpp
attributes[0].offset = 0;
attributes[0].format = VK_FORMAT_R32G32B32_SFLOAT;
attributes[1].offset = 0;
attributes[1].format = VK_FORMAT_R32G32_SFLOAT;
```
And in this case it would only read 3 floats per vertex and
```glsl
inA.x == inB.x
inA.y == inB.y;
```

abbaswasim:
> So using your example with VK_FORMAT_R32G32B32A32_SFLOAT, then the developer just needs to update where in the Vertex stride to grab the second attribute

Yes indeed but in your example you are not updating the `offset`, you are only updating the `format`. Ignore my example if that has created confusion. You need to fix the offset for second attribute `u, v`. And if you do that the then method of "Adjusting format" is redundant.

> To add to this, I am pretty sure it also legal to go

That is also true. I am not saying you can't have overlapping attributes or similar offsets. I am just saying in your example your second attribute which is suppose to be `u, v` values are not correct. They are instead both pointing at `pad` with the offset of `3 * sizeof(float);`

sfricke-samsung:
ok, fair, so the simple fix would to add a blurb about needing that offset set to `4 * float` to work...

Here, if you don't mind, let me purpose of reworking of that part of the chapter, re-reading it for this PR I realized it redundant, but at the same time, I really want to have some example to help explain the behavior of how the `format` is reading raw data

PR #90, opened 2021-03-15
Update wsi.md

Andreyogld3d:
QNX


sfricke-samsung:
Thanks for splitting it up!

PR #93, opened 2021-04-04
Remove duplicate text

RoryO:


sfricke-samsung:
thanks, not sure how that extra text snuck in there

RoryO:
> thanks, not sure how that extra text snuck in there

If you use vim or a vim like editing for prose, I've frequently made the mistake of period-esc-period ending a sentence. That takes the last chunk you entered in insert mode and inserts it again. This looked like a similar mistake!

PR #96, opened 2021-04-11
Add Push Constant chapter

sfricke-samsung:


oddhack:
@sfricke-samsung likewise this will require asciidoc conversion, which I will do, if you want.

sjfricke:
So the original goal of this chapter was to answer the last section question of `## Pushing data with non-compatible layouts`... the internal issue hit many problems and wasn't much energy to get it resolved

But looking at this, there is still some good content in here and I think it is probably worth rebasing and getting in

sjfricke:
I can asciidoctor this btw, will probably just make a new PR because this fork is on my old account

sjfricke:
@oddhack or @gpx1000 this can be closed now with https://github.com/KhronosGroup/Vulkan-Guide/pull/196 merged (I don't have access to my old account to close it myself)

PR #99, opened 2021-04-16
Added note on support check for iOS

SaschaWillems:
This PR adds a note on LunarG's port of the Vulkan Hardware Capability Viewer for iOS, which can be used to check Vulkan support similar to Android.


sfricke-samsung:
LGTM

PR #100, opened 2021-04-16
Changed wording on core versions, added some clarification

SaschaWillems:
This PR slightly rephrases the introductory chapter for the Vulkan version and adds an explanation at what promotion to core means.


sfricke-samsung:
Thanks for improving this!

PR #108, opened 2021-05-06
Update the Vulkan Layers section

christophe-lunarg:
- Add missing VK_LAYER_KHRONOS_synchronization2 documentation
- Clean up VK_LAYER_KHRONOS_validation section and its feature that was claimed to be layers but that are actually validation features
- Improve layout

Note that in July Vulkan SDK release, Vulkan Configurator will point to https://github.com/KhronosGroup/Vulkan-Guide/blob/master/chapters/development_tools.md#vulkan-layers so that Vulkan developers could find new layers and so that the Vulkan Layers ecosystem could grow.


sfricke-samsung:
lgtm

PR #117, opened 2021-05-25
Add Queue Chapter

sfricke-samsung:


chapters/queues.md
```diff
@@ -0,0 +1,100 @@
+# Queues
+
+> Further resources for queues from [AMD](https://gpuopen.com/learn/concurrent-execution-asynchronous-queues/) and [NVIDIA](https://www.khronos.org/assets/uploads/developers/library/2016-vulkan-devday-uk/9-Asynchonous-compute.pdf)
+
+A `VkQueue` is what an application submits work to, normally in the form of `VkCommandBuffer` objects or [sparse bindings](./sparse_resources.md).
```
fluppeteer, line 5: I know we're trying to define VkQueue, but allowing for highlighting, I hope "An application submits work to a 'VkQueue'..." would be clear enough without the awkward wording?

Otherwise "A 'VkQueue' is the type of object to which an application submits work", but this is the type of English with which we will not put...

chapters/queues.md
```diff
@@ -0,0 +1,100 @@
...
+
+Command buffers submitted to a `VkQueue` start in order, but are allowed to proceed independently after that and complete out of order.
+
+Command buffers submitted to different queues are unordered relative to each other unless you explicitly synchronize them with a `VkSemaphore`.
+
+You can only submit work to a `VkQueue` from one thread at a time, but different threads can submit work to different a `VkQueue` simultaneously.
+
+The concept of a `vkQueue` is implementationed defined. Some hardware have multiple hardware queues and submitting work to multiple `vkQueue` will proceed independently and concurrently in the hardware. Some implementation will do some scheduling at a kernel driver level before being submitted to the hardware. There is no current way in Vulkan to expose the exact details how each `vkQueue` is mapped to the implementation.
```
fluppeteer, line 13: implementation-defined

chapters/queues.md
```diff
@@ -0,0 +1,100 @@
...
+
+Command buffers submitted to a `VkQueue` start in order, but are allowed to proceed independently after that and complete out of order.
+
+Command buffers submitted to different queues are unordered relative to each other unless you explicitly synchronize them with a `VkSemaphore`.
+
+You can only submit work to a `VkQueue` from one thread at a time, but different threads can submit work to different a `VkQueue` simultaneously.
+
+The concept of a `vkQueue` is implementation-defined. Some hardware have multiple hardware queues and submitting work to multiple `vkQueue` will proceed independently and concurrently in the hardware. Some implementation will do some scheduling at a kernel driver level before being submitted to the hardware. There is no current way in Vulkan to expose the exact details how each `vkQueue` is mapped to the implementation.
+
+> Not all applications will require or benefit from mulitple queues. It is reasonable for an application to have a single "universal" graphics supported queue to submit all the work to the GPU.
```
janharaldfredriksen-arm, line 15:
```suggestion
> Not all applications will require or benefit from multiple queues. It is reasonable for an application to have a single "universal" graphics supported queue to submit all the work to the GPU.
```

chapters/queues.md
```diff
@@ -0,0 +1,100 @@
...
+
+An application submits work to a `VkQueue`, normally in the form of `VkCommandBuffer` objects or [sparse bindings](./sparse_resources.md).
+
+Command buffers submitted to a `VkQueue` start in order, but are allowed to proceed independently after that and complete out of order.
+
+Command buffers submitted to different queues are unordered relative to each other unless you explicitly synchronize them with a `VkSemaphore`.
+
+You can only submit work to a `VkQueue` from one thread at a time, but different threads can submit work to different a `VkQueue` simultaneously.
+
+The concept of a `vkQueue` is implementation-defined. Some hardware have multiple hardware queues and submitting work to multiple `vkQueue` will proceed independently and concurrently in the hardware. Some implementation will do some scheduling at a kernel driver level before being submitted to the hardware. There is no current way in Vulkan to expose the exact details how each `vkQueue` is mapped to the implementation.
```
janharaldfredriksen-arm, line 13: I don't think the "concept" is implementation defined. I think you're trying to say that it's implementation-defined how a queue is mapped to the underlying hardware (sort of what the last sentence of the paragraph is saying.

Also, the grammar in the second sentence is a bit broken. I think "will proceed" points back to the "submitting" in the sentence, but I suspect that what you're trying to express is that the submitted work proceeds in parallel. It would also be good to avoid using "hardware" three times in the same sentence.

Finally,
>Some implementation will do some scheduling at a kernel driver level before being submitted to the hardware.
Maybe:
=> Some implementations will do scheduling at a kernel driver level before submitting work to the hardware.

chapters/queues.md
```diff
@@ -0,0 +1,100 @@
...
+
+> Not all applications will require or benefit from mulitple queues. It is reasonable for an application to have a single "universal" graphics supported queue to submit all the work to the GPU.
+
+# Queue Family
+
+There are various types of operations a `VkQueue` can support. A "Queue Family" just describes a set of `VkQueue` that have common properties and support the same functionality, as advertised in `VkQueueFamilyProperties`.
```
janharaldfredriksen-arm, line 19: 'VkQueue' needs to be plural when describing the set of them.

chapters/queues.md
```diff
@@ -0,0 +1,100 @@
...
+        break;
+    }
+}
+```
+
+## Creating and getting a Queue
+
+Unlike other handles such as `VkDevice`, `VkBuffer`, `VkDeviceMemory`, there is **no** `vkCreateQueue` or `vkAllocateQueue`. Instead, the driver is in charge of creating and destroying the `VkQueue` handles during `vkCreateDevice`/`vkDestroyDevice` time.
+
+The following examples will use the hypothetical implementation which support 3 `VkQueue` from 2 Queue Families:
```
janharaldfredriksen-arm, line 69: `VkQueue`'s (or some other plural)

chapters/queues.md
sfricke-samsung: Agree and reworded the sentence to flow better, not repeat, and get rid of "concept"

PR #134, opened 2021-09-12
Fix VK_KHR_imageless_framebufferchapter formatting

SaschaWillems:
This PR fixes a very minor formatting error on the VK_KHR_imageless_framebuffer page, where an inline code block was not properly ended.


sfricke-samsung:
thanks for spotting this and sending patch!

PR #136, opened 2021-09-16
Update depth.adoc

sein519:
Fix the calculation expression of Zf(framebuffer depth coordinate)


sfricke-samsung:
nice catch! Thanks for fixing this

PR #142, opened 2021-09-27
Fix improper notations of VkQueue

sein519:


sfricke-samsung:
Thanks for cleaning this chapter up

PR #144, opened 2021-09-28
Add chapter on using HLSL with Vulkan

### LINKED ISSUE #122: Adding chapter on HLSL shader model coverage
As the question comes up here and there, would it make sense to have chapter on Vulkan's coverage of the different HLSL shader models?

Similar to this slide [from this presentation](https://www.khronos.org/assets/uploads/developers/library/2018-siggraph/05b-Vulkan-Shader-Compiler-Updates_Aug18.pdf):

image

E.g. for [SM 6.6](https://devblogs.microsoft.com/directx/in-the-works-hlsl-shader-model-6-6/) afaik at least some features are already present in Vulkan via certain extensions.

SaschaWillems:
This PR adds a new chapter on how to use HLSL with Vulkan. It contains some background information, syntax comparison, etc. It's meant as a general overview to let people know that HLSL is a first-class language in Vulkan.

I also added a shader model coverage table. Please note that I'm not too knowledgeable with HLSL's shader model to it may not be 100% complete or correct. I'm also not sure if we should explicitly mention extensions in there, as I did with SM6.6. But maybe we can build on this and improve the table over time.

Since this is my first chapter for the guide, I hope my stile of writing is fine and fits into the guide. I'm not a native speaker, so wording may sound odd in places and I bet there are commas in places where they don't belong ;)

Any feedback is appreciated :)

Fixes #122


README.adoc
```diff
@@ -177,6 +177,9 @@ The Vulkan Guide is designed to help developers get up and going with the world
 
 // include::{chapters}common_pitfalls.adoc[]
 
+== xref:{chapters}hlsl.adoc[Using HLSL shaders]
```
sfricke-samsung, line 180: so you also need to add this to the `guide.adoc` file (for the time being)

chapters/hlsl.adoc
```diff
@@ -0,0 +1,164 @@
...
+Vulkan does not directly consume shaders in a human-readable text format, but instead uses link:./what_is_spirv.adoc[SPIR-V] as an intermediate representation. This opens the option to use shader languages other than e.g. GLSL, as long as they can target the Vulkan SPIR-V environment.
+
+One such language is the High Level Shading Language (HLSL) by Microsoft, used by DirectX. Thanks to link:https://www.khronos.org/blog/hlsl-first-class-vulkan-shading-language[recent additions to Vulkan 1.2] it is now considered a first class shading language for Vulkan that can be used just as easily as GLSL.
+
+With link:https://github.com/microsoft/DirectXShaderCompiler/blob/master/docs/SPIR-V.rst#unsupported-hlsl-features[a few exceptions], all Vulkan features and shader stages available with GLSL can be used with HLSL too, including recent Vulkan additions like hardware accelerated ray tracing. On the other hand, HLSL to SPIR-V supports Vulkan exclusive features that are not (yet) available in DirectX.
+
+image::images/what_is_spirv_dxc.png[what_is_spriv_dxc.png]
+
+[[applications-pov]]
+= From the application's point-of-view
```
sfricke-samsung, line 19: nit: each chapter should only have one `H1` header so just add an extra `=` to the rest of the headers in this chapter

chapters/hlsl.adoc
```diff
@@ -0,0 +1,164 @@
+// Copyright 2021 The Khronos Group, Inc.
+// Copyright 2021 Sascha Willems
+// SPDX-License-Identifier: CC-BY-4.0
+
+ifndef::chapters[:chapters:]
+
+[[hlsl-in-vulkan]]
+= HLSL in Vulkan
```
sfricke-samsung, line 8: would point a link here in the "what is SPIR-V" chapter!

chapters/hlsl.adoc
```diff
@@ -0,0 +1,164 @@
...
+
+Vulkan does not directly consume shaders in a human-readable text format, but instead uses link:./what_is_spirv.adoc[SPIR-V] as an intermediate representation. This opens the option to use shader languages other than e.g. GLSL, as long as they can target the Vulkan SPIR-V environment.
```
sfricke-samsung, line 10:
```suggestion
Vulkan does not directly consume shaders in a human-readable text format, but instead uses xref:{chapters}what_is_spirv.adoc[SPIR-V] as an intermediate representation. This opens the option to use shader languages other than e.g. GLSL, as long as they can target the Vulkan SPIR-V environment.
```

sfricke-samsung (changes requested):
This is great and thanks for writing this up!

Just a few nit, otherwise LGTM!

chapters/hlsl.adoc
sfricke-samsung: same for the other references to other chapters

README.adoc
SaschaWillems: I see. That's the "whole guide in one place" document. I have added the new chapter in there too.

chapters/hlsl.adoc
SaschaWillems: Updated this and the other link to internal files and replaced them with xref as suggested.

chapters/hlsl.adoc
SaschaWillems: I added a link to the first occurrence of HLSL, but it doesn't seem to work.

Code looks like this:

```
xref:{chapters}hlsl.adoc[HLSL]
```

But clicking on it results in an URl ending with `chapters/%7Bchapters%7Dhlsl.adoc` that doesn't work. Any idea? Is the link format wrong? That same format works fine in my new chapter.

sfricke-samsung: awww, the spir-v chapter is missing

```
ifndef::chapters[:chapters:]
```
let me add this for any other missing chapters to

So just leave (or rebase if you want) because when the merge happens from this PR it will all just work

README.adoc
sfricke-samsung: ya, if you run the `make` it builds a cool single-page guide, the goal is in the future to host that more properly somewhere if we would want

sfricke-samsung (approved):
Will keep up for a few days to give others a chance to review too, but otherwise everything looks good to me now

Andreyogld3d (commented):
I suggest we use Direct3D instead of DirectX, it would be technically correct

SaschaWillems:
I used Direct3D in a first draft, but even Microsoft refers to DirectX:

"HLSL is the C-like high-level shader language that you use with programmable shaders in DirectX." (from the official MS page for HLSL)

PR #161, opened 2022-01-21
Add dynamic offset chapter

sfricke-samsung:
Written as part of the solution to https://github.com/KhronosGroup/Vulkan-Docs/issues/1546


chapters/descriptor_dynamic_offset.adoc
```diff
@@ -0,0 +1,129 @@
...
+    1,              // descriptorSetCount,
+    &descriptorSet, // pDescriptorSets,
+    1,              // dynamicOffsetCount
+    offsets         // pDynamicOffsets
+);
+----
+
+This is what it looks like with the invalid dynamic offset
+
+image::images/descriptor_dynamic_offset_example_d.png[descriptor_dynamic_offset_example_d.png]
```
cforfang, line 125: In the image, the start of the range is offset=4, but seems it should be offset=12 (4 static + 8 dynamic)?

sfricke-samsung: good catch, I see how this happen, will fix it asap

PR #171, opened 2022-03-26
Add Vulkan shader stage to HLSL target shader profile mapping table

SaschaWillems:
This PR adds a a new table to the HLSL chapter that maps Vulkan shader stages to HLSL target shader profiles. This table can be used to look up the target shader profile name required for compiling a given HLSL shader to SPIR-V.


sfricke-samsung:
Thanks for adding the table (and probably having to look up what asciidoc table syntax is :sweat_smile: at least I always have too)

PR #184, opened 2022-06-19
Add in the Japanese translation from Takashi Umezawa and Nishidate-san.

gpx1000:
Moved the translation into a lang directory with the expectation that other language translations might be able to go there in the future.


lang/jp/README-jp.adoc
```diff
@@ -0,0 +1,299 @@
...
+:regtitle: pass:q,r[^®^]
+The Khronos{regtitle} Vulkan Working Group
+:data-uri:
+:icons: font
+:max-width: 100%
+:numbered:
+:source-highlighter: rouge
+:rouge-style: github
+
+image::../images/vulkan_logo.png[Vulkan Logo]
```
sjfricke, line 14: the image links are going to need to be adjusted

gpx1000: Thanks for the catch Spencer!  I think these were the last two that were not adjusted correctly as the README-jp.adoc isn't processed by adoctor when building the guide (i.e. the rest spit out errors).
I'll resolve the conversation for now, please let me know if you find anything else.
I don't know if you're able but Marty has us scheduled to discuss the guide in the samples meeting tomorrow.

README.adoc
```diff
@@ -14,6 +14,11 @@ The Khronos{regtitle} Vulkan Working Group
 image::images/vulkan_logo.png[Vulkan Logo]
 image::images/khronos_logo.png[Khronos logo]
 
+:lang-jp: lang/jp/
+This document is also available in the following languages: +
+xref:{lang-jp}README-jp.adoc:[Japanese]
```
sjfricke, line 19: one more quick fix

```suggestion
xref:{lang-jp}README-jp.adoc[Japanese]
```

resolves to `lang/jp/README-jp.adoc:` not `lang/jp/README-jp.adoc`

lang/jp/chapters/enabling_extensions.adoc
```diff
@@ -0,0 +1,101 @@
...
+
+拡張機能には、**インスタンス拡張機能**と**デバイス拡張機能**の2つのグループがあります。簡単に言えば、**インスタンス拡張機能** は `VkInstance` 全体に結び付けられ、**デバイス拡張機能** は単一の `VkDevice` インスタンスにのみ結び付けられます。
+
+この情報は、各拡張機能のリファレンスページの「`Extension Type`」セクションに記載されています。以下に例を示します。
+
+image::../../../chapters/images/enabling_extensions_instance_extension.png[enabling_extensions_instance_extension.png]
+
+== 対応の確認
+
+アプリケーションは、まずlink:https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#extendingvulkan-extensions[物理デバイスにクエリ]し、`vkEnumerateInstanceExtensionProperties` または `vkEnumerateDeviceExtensionProperties` で拡張機能が**サポート**されているかどうかを確認することができます。
```
sjfricke, line 21: もし品題で`対応`を使ったら、どうしてここで`対応`のかわりに`サポート`を使った？

@gpx1000 maybe this brings up a good future logistical question of if someone else (who's Japanese is better than mine) has a question about the translation who is able to respond to such questions

gpx1000: You're asking the right questions...

This is likely to be a problem, however, I seem to remember @marty-johnson59 mentioning that Khronos might have someone on staff.  I think the process will be we bring up all issues/questions to the samples working group and from there we'll route to hopefully someone who can help.

gpx1000: If you used correspondence in your article, why did you use support instead of correspondence here? <-- From google translate this is the question.  I wonder if logistically it would make sense to run through Google translate the questions / replies and have a native speaker double check?

sjfricke: So without going deep into Japanese, they translated "support" as カタカナ and 漢字 and not sure which is normally used more often, but surprised it wasn't consistent in both spots...also Google translate is an amazing tool, but not for everything such as technical documentation

gpx1000: I agree, which is really the reasoning for having a request for support from a native speaker to handle the translations.
For the broader point of how to support questions dynamically, best idea I have is to provide best efforts via bringing it back to the WG, looking for resources within Khronos and member companies to find someone capable of addressing the issues as they come up.  I am wholly unqualified to even pretend to translate Japanese let alone any other language so I'd be unable to really do much more than that.

TakashiKhronos: ah, in case of support, using サポート or 対応 is not a big issue for understanding the context since both does mean the same thing in Japanese. So, it's not a big issue, and no Japanese would care about the difference between サポート and 対応 in a sentence.
Could you please accept as is rather than taking more time to discuss here? We can modify/update if there is serious translation mismatch. Thanks!

sjfricke: I agree, rather have something tangible out and can always adjust, (cc @marty-johnson59 )

PR #191, opened 2022-08-04
Fix the CI delay timeout.

gpx1000:
The Apple Appstore awesome_bot web check is causing issues in the CI.  Fixing it two ways, increasing the timeout allowance and adding the Apple Appstore to the white list.

We should consider if it's worth adding --allow-timeout to the awesome_bot CI check and dropping the timeout to 5 for CI speed improvement.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -32,4 +32,4 @@ jobs:
       - name: awesome_bot
         run: |
           gem install awesome_bot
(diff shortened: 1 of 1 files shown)


sjfricke (approved):
> We should consider if it's worth adding --allow-timeout to the awesome_bot CI check and dropping the timeout to 5 for CI speed improvement.

I guess I never thought of CI Speed for the guide before as urgent and why never touched the timeout value before

marty-johnson59:
yah, it's MUCH faster than building samples... But, couldn't hurt (I hope).  Merged..

PR #226, opened 2023-07-22
Add helper map for Dynamic State

spencer-lunarg:
After spending a LOT of time having to bounce around the spec to find information about each dynamic state while working on the Validation Layers, I decided it will be worth my time in the future to have this mapping of everything

Will play around later with the width to see if can remove the scrolling (no asciidoc pro), but figured I am going to `ctrl+f` mostly anyways

(cc @pdaniell-nv @ziga-lunarg @daniel-story)


yknishidate:
Hi! thanks for adding Japanese too! It is correct that the Japanese translation of map is 地図. However, since it usually has a physical/geopolitical meaning, it looks a bit unnatural to Japanese speakers. In this case, it is better to translate it as マップ in katakana.

spencer-lunarg:
@yknishidate ありがとう、I wasn't sure, thats why I put in in `「」`, but will switch the `マップ`

gpx1000 (approved):

PR #227, opened 2023-08-03
Fix warning about existing reference for push-constants id. 

reneruhr:
Fixed link to push-constant section.

Original warning when running `make`:
`asciidoctor: WARNING: chapters/push_constants.adoc: line 8: id assigned to section already in use: push-constants`


spencer-lunarg (approved):
LGTM, thanks for catching this, I realize I still need to get CI up to catch these earlier

PR #236, opened 2023-10-16
Fix a formatting issue caused by a non-escaped apostrophe

SaschaWillems:
This PR fixes a minor formatting issue caused by an apostrophe that wasn't properly escaped in inline code. That caused parts of the non-code text to be formatted as code too.


spencer-lunarg:
image

wow thanks!....I've tried looking into this before and basically wanted this lol

There are some other places, but I can get them

spencer-lunarg (approved):

spencer-lunarg:
nvm the other issue was 

> `VkQueue`&#8203;s

which is slightly different

PR #246, opened 2024-01-24
Fix Broken Vulkan-Samples link

hyperlogic:
Changed from .md to .adoc


spencer-lunarg (approved):
Thanks for fixing this!

PR #266, opened 2024-07-15
[JP] Add shader comparison page

yknishidate:
Added Japanese version of high level shader lang comparison page and made minor corrections to the original page.


spencer-lunarg (approved):
ありがとう！

PR #277, opened 2024-09-17
Fix wrong link on spirv_extensions.adoc

burak-efe:
Link was pointing to "64 bit atomics" extension instead of "8 bit storage".


spencer-lunarg (approved):

spencer-lunarg:
thanks for making the patch!

PR #285, opened 2024-12-10
Hide logos for docs site build, fix comment format

SaschaWillems:
With https://github.com/KhronosGroup/antora-ui-khronos/pull/11 we just merged dark mode for the docs site. This has issues with images that use black as a fixed color:

image

As we don't display logos in the other components of the docs site, this PR simply hides the logos when building for the docs site.

It also fixes the comments being wrongly formatted, so they no longer show up as text on the page.

Would be great if we could merge this before Jon releases the next docs site update. Otherwise the guide landing page would look odd.


gpx1000 (approved):

spencer-lunarg (approved):
much simpler fix than I was expecting from the description :laughing:

Issue #287, opened 2025-01-06
[BUG] Links to the specification are broken


skibon02:
**Describe the bug**
The page https://docs.vulkan.org/guide/latest/vulkan_spec.html lists some links to various specification versions and formats, but most of them are broken.
1. All links under "Prebuilt HTML chunked vulkan spec" lead to non-existing resources.
2. When index.html is manually added to end of all links (for example https://registry.khronos.org/vulkan/specs/1.3/html/index.html) they are redirecting to https://docs.vulkan.org/spec/latest/chapters/introduction.html instead of prebuilt specification document
3. All links under "Prebuilt HTML Full Vulkan Spec" lead to the latest specification version with all extensions.
4. All links under "Prebuilt PDF Vulkan Spec" lead to the latest specification in pdf format with all extensions.

**To Reproduce**
1. Go to https://docs.vulkan.org/guide/latest/vulkan_spec.html
2. Click on "Prebuilt HTML Chunked Vulkan Spec" -> "Vulkan 1.0 Specification" -> Core
3. Observe not found resource in Khronos Registry

**Expected behavior**
The specification is expected to open in the chosen format

**Screenshots**
![image](https://github.com/user-attachments/assets/7ced35aa-90ba-4a57-8791-e9f60d78c8c2)

**Desktop (please complete the following information):**
 - OS: Linux Mint
 - Browser Brave
 - Version 1.73.104

spencer-lunarg:
I can update these

PR #288, opened 2025-01-17
Cleanup Vulkan Spec chapter

spencer-lunarg:
closes https://github.com/KhronosGroup/Vulkan-Guide/issues/287

With both the new antora site and using `latest`, this chapter needed a cleanup with the goal to keep things simpler (I felt we now can remove a lot of legacy knowledge and aim people towards the new modern flows we have)


SaschaWillems (approved):
LGTM 👍🏻

PR #292, opened 2025-02-10
Add Location Component chapter

spencer-lunarg:
@Tobski @alan-baker - Here is a simple chapter around our favorite topic of `Location`/`Component`

1. Please review and make sure we all agree
2. Please suggest other possible "fun" cases to add here for help in future references


chapters/location_component_interface.adoc
```diff
@@ -0,0 +1,98 @@
...
+
+== Basic Example
+
+Here is a basic example:
+
+[source,glsl]
+----
+layout(location=0) in vec4 a;
+layout(location=1, component = 0) in vec2 b;
+layout(location=1, component = 2) in float c;
```
Tobski, line 24: Somewhere this could do with an explicit call-out that the vec2 consumes two components.

chapters/location_component_interface.adoc
```diff
@@ -0,0 +1,98 @@
...
+layout(location=0, component=1) in float in_f;
+----
+
+== Array
+
+Arrays will consume the entire `Location` for a given element in the array, as an example
+
+[source,glsl]
+----
+layout(location=0, component=1) in float a[3];
```
Tobski, line 78:
```suggestion
layout(location=0) in float a[3];
```

component decoration should just be banned (even if it's technically not atm)

chapters/location_component_interface.adoc
```diff
@@ -0,0 +1,98 @@
...
+
+image::{images}location_example_array.svg[location_example_array]
+
+As seen, using a scalar or something such as a `vec2`/`float2` will leave many `Component` slots unused.
+
+It is **not** allowed to use any other `Component` in a `Location` that is being consumed by an array
+
+[source,glsl]
+----
+layout(location=0, component=1) in float a[3];
```
Tobski, line 89:
```suggestion
layout(location=0) in float a[3];
```

chapters/location_component_interface.adoc
```diff
@@ -0,0 +1,98 @@
...
+layout(location=2, component=2) in float b;
```
Tobski, line 90: A comment here explaining would be helpful - could also be a sentence afterwards.

```suggestion
layout(location=2, component=2) in float b;  // Invalid, as the array consumes all of location 2.
```

chapters/location_component_interface.adoc
```diff
@@ -0,0 +1,98 @@
...
+
+[source,glsl]
+----
+layout(location=0) in vec2 in_a;
+layout(location=0, component=1) in float in_f;
+----
+
+== Array
+
+Arrays will consume the entire `Location` for a given element in the array, as an example
```
Tobski, line 74: This should explain that each subsequent element consumes more locations. Maybe something like:

 ```suggestion
An element of an array will consume all every `Component` in a `Location` that it would consume as a non-arrayed value, with each subsequent element consuming the next available `Location`. 
```

**Review by Tobski — CHANGES_REQUESTED** [review]
Some comments for clarity and a few fixes. I think it would also be a good idea to have a 64-bit example, particularly a 64-bit vec3/4, showing how these consume multiple locations, and how these have to be aligned to even components.

File: chapters/location_component_interface.adoc [hunk 6]
```diff
@@ -0,0 +1,98 @@
...
+
+image::{images}location_example_boundaries.svg[location_example_boundaries]
+
+== 64-bit
+
+64-bit are special as they can consume up to 2 `Location`, but they must only start at `Component` `0` or `2`.
+
+[source,glsl]
+----
+layout(location=0) in f64vec3 a;
```
spencer-lunarg, line 57: >  I think it would also be a good idea to have a 64-bit example, particularly a 64-bit vec3/4

Here is that example

chapters/location_component_interface.adoc
```diff
@@ -0,0 +1,146 @@
...
+ifndef::images[:images: images/]
+
+[[location-and-component-interface]]
+= Location and Component Interface
+
+This chapter is an overview of the link:https://docs.vulkan.org/spec/latest/chapters/interfaces.html#interfaces-iointerfaces-locations[Location and Component Assignment] chapter to help give examples, especially around some of the more extreme edge cases.
+
+The simplest way to think about a `Location` is that it is made up of four 32-bit `Components`.
+This means a `vec4`/`float4`/`uvec4`/etc will fit perfectly in a single `Location`.
+Multiple variables can be packed into the same `Location` if they are specified for unique components.
```
alan-baker, line 14: Maybe it's better to reuse the Vulkan spec language here as an aid when reading that chapter.
```suggestion
Multiple variables can be packed into the same `Location` if their `Component`s do not overlap.
```

chapters/location_component_interface.adoc
```diff
@@ -0,0 +1,146 @@
...
+image::{images}location_example_64bit.svg[location_example_64bit]
+
+== Interleaving Components
+
+The following attempt to have multiple variables alias the same component is link:https://godbolt.org/z/h61baYhT4[not allowed].
+
+[source,glsl]
+----
+layout(location=0) in vec2 in_a;
+layout(location=0, component=1) in float in_f;
```
alan-baker, line 72: Possible improvement to be more explicit about what's happening:
```suggestion
layout(location=0) in vec2 in_a; // Components 0 and 1
layout(location=0, component=1) in float in_f; // Invalid: overlaps component 1
```

chapters/location_component_interface.adoc
```diff
@@ -0,0 +1,146 @@
...
+----
+
+The following modification would make it legal as multiple variable can share a `Location`, just not a `Component
+`
+[source,patch]
+----
+layout(location=0) in vec2 in_a;
+- layout(location=0, component=1) in float in_f;
++ layout(location=0, component=2) in float in_f;
```
alan-baker, line 81:
```suggestion
layout(location=0) in vec2 in_a; // Components 0 and 1
// Change in_f to use component 2 instead
- layout(location=0, component=1) in float in_f;
+ layout(location=0, component=2) in float in_f;
```

chapters/location_component_interface.adoc
```diff
@@ -0,0 +1,146 @@
...
+----
+
+== Array
```
alan-baker, line 84: Once Vulkan issue 4180 is clarified it would be good to expand this section I think

alan-baker (commented):
Should this chapter provide guidance on whether the number of locations ought to be minimized? Maybe it should also mention that you can't mix interpolation within a location?

spencer-lunarg:
> Should this chapter provide guidance on whether the number of locations ought to be minimized? Maybe it should also mention that you can't mix interpolation within a location?

1. Yes, can add a thing at the top explaining the desire to use less if possible (and also there are limit values)
2. I forgot about the interpolation case, will add that

spencer-lunarg:
@alan-baker for the 

>  that you can't mix interpolation within a location?

Did you mean between two stages? (I plan to have another possible guide about connecting the Location)

As for what I think you mean, it seems that GLSL disallows it, but not `spirv-val` (https://godbolt.org/z/EohPberKd)

alan-baker:
> @alan-baker for the
> 
> > that you can't mix interpolation within a location?
> 
> Did you mean between two stages? (I plan to have another possible guide about connecting the Location)
> 
> As for what I think you mean, it seems that GLSL disallows it, but not `spirv-val` (https://godbolt.org/z/EohPberKd)

Yes, I know spirv-val doesn't check it. It's good that glslang does, but it's one of those things for new frontends that is sort of a gotcha that's not easy to find.

spencer-lunarg:
@Tobski @alan-baker if you are ok with the current state of this?

For the Interpolation (and other related checks) happy to have a follow up, if this is "correct" it is IMO better then having nothing

Honestly this is really just for someone like me who has to re-reference this stuff once a year and needs a quick reminder what we agreed on

Tobski (approved):

alan-baker (approved):

PR #293, opened 2025-02-16
Keep spec links in Antora

spencer-lunarg:
Went through and replaced old `registry.khronos.org` links to `docs.vulkan.org` to keep people inside the Antora site as much as possible

(small few fixes found along the way mended as well)


SaschaWillems (approved):
LGTM!

PR #299, opened 2025-03-27
Add link to find Dynamic State Map

spencer-lunarg:
https://docs.vulkan.org/guide/latest/dynamic_state.html doesn't point to https://docs.vulkan.org/guide/latest/dynamic_state_map.html as pointed out in https://github.com/KhronosGroup/Vulkan-Docs/issues/2504#issuecomment-2718671498

### DIFF
--- a/chapters/dynamic_state.adoc
+++ b/chapters/dynamic_state.adoc
@@ -9,7 +9,7 @@ ifndef::images[:images: images/]
 
 [NOTE]
 ====
-link:https://docs.vulkan.org/spec/latest/chapters/pipelines.html#pipelines-dynamic-state[Vulkan Spec chapter]
+xref:{chapters}dynamic_state_map.adoc[List of all dynamic states with details]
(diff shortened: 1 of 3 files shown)


chapters/dynamic_state.adoc
```diff
@@ -9,7 +9,7 @@ ifndef::images[:images: images/]
 
 [NOTE]
 ====
-link:https://docs.vulkan.org/spec/latest/chapters/pipelines.html#pipelines-dynamic-state[Vulkan Spec chapter]
+Additional resources: xref:{chapters}dynamic_state_map.adoc[List of all dynamic states with details]
 ====
 
 == Overview
```
spencer-lunarg, line 15: In here we already point to `pipelines.html#pipelines-dynamic-state` so no need to have the link twice at the top-ish

SaschaWillems (approved):
LGTM!

PR #301, opened 2025-04-17
Compiler Explorer example for VK_KHR_storage_buffer_storage_class

spencer-lunarg:
has to show yet-another-person what `VK_KHR_storage_buffer_storage_class` was and why using glslang with `--target-env vulkan1.0` is bad, so decided to put the compiler explorer example here


SaschaWillems (approved):
LGTM :)

PR #308, opened 2025-05-02
Add Swapchain Semaphore Reuse chapter

spencer-lunarg:
Thanks to @artem-lunarg for writing this up, this will likely be very popular as the 1.4.313 SDK throws errors for people that look like (pasting for SEO)

```
Validation Error: [ VUID-vkQueueSubmit-pSignalSemaphores-00067 ] | MessageID = 0x539277af
vkQueueSubmit(): pSubmits[0].pSignalSemaphores[0] (VkSemaphore 0x40000000004) is being signaled by VkQueue 0x5c1acf5ae3f0, but it may still be in use by VkSwapchainKHR 0x80000000008.

Here are the last 3 acquired image indices: [0], 1, 2.
(brackets mark the last use of VkSemaphore 0x40000000004 in a presentation operation)
Swapchain image 0 was presented but was not re-acquired, so VkSemaphore 0x40000000004 may still be in use and cannot be safely reused with image index 2.

Vulkan insight: One solution is to assign each image its own semaphore. Here are common methods to ensure that a semaphore passed to vkQueuePresentKHR is not in use and can be safely reused:
    a) Use a separate semaphore per swapchain image. Index these semaphores using the index of the acquired image.
    b) Consider the VK_EXT_swapchain_maintenance1 extension. It allows using a VkFence with the presentation operation.
The Vulkan spec states: Each binary semaphore element of the pSignalSemaphores member of any element of pSubmits must be unsignaled when the semaphore signal operation it defines is executed on the device (https://docs.vulkan.org/spec/latest/chapters/cmdbuffers.html#VUID-vkQueueSubmit-pSignalSemaphores-00067)
Objects: 2
    [0] VkSemaphore 0x40000000004
    [1] VkQueue 0x5c1acf5ae3f0
```

### DIFF
--- a/chapters/swapchain_semaphore_reuse.adoc
+++ b/chapters/swapchain_semaphore_reuse.adoc
@@ -1,0 +1,140 @@
+// Copyright 2025 The Khronos Group, Inc.
+// SPDX-License-Identifier: CC-BY-4.0
+
+// Required for both single-page and combined guide xrefs to work
+ifndef::chapters[:chapters:]
+ifndef::images[:images: images/]
+
+[[swapchain-semaphore-reuse]]
+= Swapchain Semaphore Reuse
+
+It is very easy to misuse the `vkQueuePresentKHR` wait semaphore, but luckily it's also usually easy to fix things by allocating one "submit finished" semaphore per swapchain image instead of per in-flight frame.
+
+== Problem Statement
+
+This chapter discusses one way to safely reuse `vkQueuePresentKHR` wait semaphores (the ones specified in `VkPresentInfoKHR::pWaitSemaphores`). In this context, *safely* means that the Vulkan specification guarantees the semaphore is no longer in use and can be reused. Since Vulkan SDK 1.4.313, the validation layer reports cases where the present wait semaphore is not used safely.
+
+[NOTE]
+====
+This is currently reported as `VUID-vkQueueSubmit-pSignalSemaphores-00067` or you may see "your VkSemaphore is being signaled by VkQueue, but it may still be in use by VkSwapchainKHR"
+====
+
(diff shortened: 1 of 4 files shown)


chapters/swapchain_semaphore_reuse.adoc
```diff
@@ -0,0 +1,140 @@
...
+== Problem Statement
+
+This chapter discusses one way to safely reuse `vkQueuePresentKHR` wait semaphores (the ones specified in `VkPresentInfoKHR::pWaitSemaphores`). In this context, *safely* means that the Vulkan specification guarantees the semaphore is no longer in use and can be reused. Since Vulkan SDK 1.4.313, the validation layer reports cases where the present wait semaphore is not used safely.
+
+[NOTE]
+====
+This is currently reported as `VUID-vkQueueSubmit-pSignalSemaphores-00067` or you may see "your VkSemaphore is being signaled by VkQueue, but it may still be in use by VkSwapchainKHR"
+====
+
+First, let's see if there's something special about present wait semaphores. The essence of Vulkan synchronization and synchronization in general is ordering. In Vulkan, this is done by *waiting* for signals, polling, etc. The entity we wait for must signal a status change or provide a query mechanism. For example, `vkQueueSubmit` allows you to specify a fence that will be signaled when workload is done and the application can wait for this signal on the CPU (host) side. `vkQueueSubmit` can also signal semaphores which are waitable only on the GPU (device) side.
```
artem-lunarg, line 22: I forgot that timeline semaphores are waitable on the host too. Could you update the ending to something like:
"`vkQueueSubmit` can also signal semaphores which are waitable both on the GPU (device) side and on the host, if they are timeline semaphores".

spencer-lunarg:
```suggestion
First, let's see if there's something special about present wait semaphores. The essence of Vulkan synchronization and synchronization in general is ordering. In Vulkan, this is done by *waiting* for signals, polling, etc. The entity we wait for must signal a status change or provide a query mechanism. For example, `vkQueueSubmit` allows you to specify a fence that will be signaled when workload is done and the application can wait for this signal on the CPU (host) side. `vkQueueSubmit` can also signal semaphores which are waitable only on the GPU (device) side (or host if using timeline semaphores).
```

thoughts?

artem-lunarg: This looks good

PR #313, opened 2025-05-11
Update EXT to VK_KHR_robustness2

spencer-lunarg:
Update EXT to KHR


spencer-lunarg:
(simple editorial fix)

PR #326, opened 2025-07-10
Extra edge case for Storage Texel Buffer

spencer-lunarg:
Was using this awesome page today to explain storage texel buffers to someone and they made a mistake (which I swear I have done 2 or 3 times over the years) and not realizing that a R32 format only get a single component written and that GLSL has a strange quirk that it requires a `gvec4`


gpx1000:
I'll take a look and see what I can do to add that info somewhere.

spencer-lunarg:
@gpx1000 realize you might be viewing from a phone, but this is a PR, not an issue

gpx1000:
Ahhh yeah ok I'm reading while charging phone and over email.  I'll look
when i get home

gpx1000 (approved):

Issue #331, opened 2025-07-21
Fix incorrect information about Texel Buffer formats


spencer-lunarg:
@gpx1000 so I learned https://github.com/KhronosGroup/Vulkan-Guide/blob/main/chapters/storage_image_and_texel_buffers.adoc#format-compatibility-requirements is wrong

See https://gitlab.khronos.org/vulkan/vulkan/-/issues/4363#note_547796

Basically I have been trying to narrow down this whole format restriction between storage images and texel buffers

Currently in Validaiton you will get a `Undefined-Value-StorageImage-FormatMismatch-ImageView` if the formats are not the exact same (https://gitlab.khronos.org/vulkan/vulkan/-/issues/4128#note_520871) and I will be soon adding the same for Texel Buffer

I have learned that a lot of this is because in OpenGL you could have things be different (like R8 and R32) and it was just "undefined values" and not "undefined behavior" (it wouldn't crash), so we need to keep things consistent it seems....

I have a new note being added to the spec to help make this a bit more clear https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/7543

PR #332, opened 2025-07-21
fix format compatibility rules for storage images

gpx1000:
addresses: #331


spencer-lunarg (approved):
thanks for the quick update... ya I was surprised to learn this whole edge case is "valid" but "undefined value" as well

PR #333, opened 2025-07-21
Add in the shader authoring guide chapter.

gpx1000:


guide.adoc
```diff
@@ -145,6 +145,8 @@ include::{chapters}common_pitfalls.adoc[]
 
 include::{chapters}hlsl.adoc[]
 
+include::{chapters}shader_authoring_guide_slang_hlsl.adoc[]
```
spencer-lunarg, line 148: missing in README.md

spencer-lunarg:
I know @SaschaWillems already wrote a whole chapter around this "area" so want him approval

Also @swoods-nv would also be great to review

chapters/shader_authoring_guide_slang_hlsl.adoc
```diff
@@ -0,0 +1,1610 @@
...
+
+While Vulkan consumes shaders in xref:{chapters}what_is_spirv.adoc[SPIR-V] format, developers can author shaders in high-level languages like HLSL (High-Level Shading Language) and Slang. This chapter builds upon the xref:{chapters}hlsl.adoc[HLSL in Vulkan] chapter, focusing specifically on authoring shaders for Vulkan applications.
+
+=== What is Slang?
+
+link:https://github.com/shader-slang/slang[Slang] is a modern shading language and compiler designed to extend HLSL with advanced features. It's backward compatible with HLSL but adds features like generics, interfaces, and reflection capabilities. Slang can target multiple backends, including SPIR-V for Vulkan.
+
+Key features of Slang include:
+
+* Full compatibility with HLSL syntax
```
swoods-nv, line 23: As above -- this is too strong a statement.

chapters/shader_authoring_guide_slang_hlsl.adoc
```diff
@@ -0,0 +1,1610 @@
...
+
+// Accessing resources
+Texture2D albedoMap = lightingParams.albedoMap;
+----
+
+This provides better organization and more flexible binding than HLSL's register-based approach.
+
+=== Syntax Differences
+
+While Slang maintains HLSL syntax compatibility, it introduces some new syntax elements:
```
swoods-nv, line 151: Same comment on "compatibility" caveats.

swoods-nv (commented):
My main overarching comment is that I'd recommend caution when phrasing migration from HLSL->Slang as "upgrading" or similar. Vulkan consciously chooses not to have an official shading language, and both Slang and HLSL continue to evolve and gain new features.

SaschaWillems:
There is def. some duplication here. The chapters I did are:

- https://github.com/KhronosGroup/Vulkan-Guide/blob/main/chapters/hlsl.adoc
- https://github.com/KhronosGroup/Vulkan-Guide/blob/main/chapters/high_level_shader_language_comparison.adoc

The latter was written with additional shading languages in mind.

So maybe it's better to

- To a general slang chapter like the one I did for HLSL (first link)
- And add slang to the already available high level shader language comparison (second link)

A separate chapter that kinda combines what we have + slang might be confusing.

SaschaWillems:
Btw. the official slang docs do have migration guides, so maybe something we could link to:

- https://docs.shader-slang.org/en/latest/coming-from-glsl.html
- https://docs.shader-slang.org/en/latest/coming-from-hlsl.html

gpx1000:
Good call on using those links, I'll add it in.  The next version will have a lot of copy pasting to place the details in the right place.  Should have a new version once I finish reviewing all changes.

gpx1000:
Alright, significant copy/pasting and reorganizing into the existing structure.  I *don't* think I lost any of the details I wrote in the original version.  However, hopefully that's closer to where things should go.

spencer-lunarg (dismissed):
- Adding Slang to the `high_level_shader_language_comparison.adoc` page is great (I often know a term in in GLSL and need to look up the Slang equivalent)
- Will wait for @SaschaWillems approval on the HLSL additions added
- Will wait for @swoods-nv  for a re-read and approval on the new Slang chapter

chapters/hlsl.adoc
```diff
@@ -7,6 +7,7 @@ ifndef::images[:images: images/]
 
 [[hlsl-in-vulkan]]
 = HLSL in Vulkan
+:toc:
```
SaschaWillems, line 10: Any reason this was explicitly added? Antora already takes care o the toc, this might lead to a duplicated toc in the docs site build.

chapters/hlsl.adoc
```diff
@@ -21,11 +22,307 @@ image::{images}what_is_spirv_dxc.png[what_is_spriv_dxc.png]
...
+    [[vk::location(2)]] float2 TexCoord : TEXCOORD0;
+};
+// Uniform buffer with transformation matrices
+struct SceneUBO {
+    float4x4 model;
+    float4x4 view;
+    float4x4 projection;
+};
+[[vk::binding(0, 0)]]
+ConstantBuffer<SceneUBO> ubo;
```
SaschaWillems, line 71: If this is meant to show a portable shader, register notation should be added.

SaschaWillems:
That's a lot of text to review ^^

Added a few comments. One of the things that IMO needs more attention is Skang's ability to automatically generate binding slots. For me, that was one of the best things about Slang, and while you can somehow find that information in this PR I'd like to see that more explicitly stated.

chapters/slang.adoc
```diff
@@ -0,0 +1,737 @@
...
+link:https://github.com/shader-slang/slang[Slang] is a modern shading language and compiler designed to extend xref:{chapters}hlsl.adoc[HLSL] with advanced features. It's largely backward-compatible with HLSL 2018 but adds features like generics, interfaces, and reflection capabilities. Slang can target multiple backends, including SPIR-V for Vulkan.
+
+With link:https://github.com/microsoft/DirectXShaderCompiler/blob/main/docs/SPIR-V.rst#unsupported-hlsl-features[a few exceptions], all Vulkan features and shader stages available with GLSL can be used with Slang too, including recent Vulkan additions like hardware accelerated ray tracing. On the other hand, Slang to SPIR-V supports Vulkan exclusive features that are not (yet) available in DirectX.
+
+image::{images}what_is_spirv_dxc.png[what_is_spriv_dxc.png]
+
+[[educational-resources]]
+== Educational resources
+
+For Slang, the official link:https://github.com/shader-slang/slang[GitHub repository] and link:https://docs.shader-slang.org/en/latest/[documentation] are the best resources to get started.
```
swoods-nv, line 23: Not sure if it's worth mentioning here, but the Slang Discord is an additional resource; this can be especially helpful if you are troubleshooting.

chapters/slang.adoc
```diff
@@ -0,0 +1,737 @@
...
+=== Offline compilation using the stand-alone compiler
+
+Compiling a Slang shader offline is straightforward:
+
+[source]
+----
+slangc -profile glsl_spirv -entry main -o output.spv input.slang
+----
+
+For HLSL compatibility mode:
```
swoods-nv, line 457: I believe there's also a GLSL full-compatibility flag (but can't remember the specific cases that it's required for -- in general, Slang can handle most GLSL without the flag. The migration Slang doc page might have more info.)

chapters/high_level_shader_language_comparison.adoc
```diff
@@ -168,13 +414,87 @@ VSOutput main(VSInput input)
...
+
+    struct Constants {
+        float4 baseColor;
+        float roughness;
+        float metallic;
+    } constants;
+};
+
+// Declare a parameter block with explicit binding
+[[vk::binding(0, 0)]]
```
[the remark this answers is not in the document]
gpx1000: Yea, you *can* specify explicit bindings but they aren't necessary as they are automatic.  The advantage to explicit is you can ensure it is assigned correctly. I took it out.

chapters/slang.adoc
gpx1000: https://docs.shader-slang.org/en/latest/coming-from-glsl.html  I don't see anything directly in this.  I did find -lang GLSL and more details in a slide show from a Vulkan SIGGRAPH BOF.  I'll add details here.

chapters/high_level_shader_language_comparison.adoc
```diff
@@ -56,15 +72,89 @@ HLSL:
...
+
+==== Generics
+
+Slang provides full generic support:
+
+[source,slang]
+----
+// Generic function
+generic<T>
+T min(T a, T b) {
```
juliusikkala: This and a couple other cases here appear to be using an old version of the generic syntax. I think this should work:
```slang
T min<T>(T a, T b) {
    return a < b ? a : b;
}
```

chapters/high_level_shader_language_comparison.adoc
```diff
@@ -56,15 +72,89 @@ HLSL:
...
+    return a < b ? a : b;
+}
+
+// Generic struct
+generic<T, let N : int>
+struct Array {
```
juliusikkala: Here too, `struct Array<T, let N : int> {` should work.

chapters/slang.adoc
gpx1000: Thanks, I just updated it to use -allow-glsl

chapters/high_level_shader_language_comparison.adoc
```diff
@@ -1014,5 +1542,61 @@ Most GLSL functions are also available in HLSL and vice-versa. This chapter list
...
 * link:https://registry.khronos.org/OpenGL-Refpages/gl4/index.php[OpenGL reference pages]
+* link:https://github.com/shader-slang/spec[Slang Language Reference]
+
+Here's an example of Slang's generic functions:
+
+[source,slang]
+----
+// Generic interpolation function
+generic<T>
+T interpolate(T a, T b, float t)
```
juliusikkala: Could be `T interpolate<T>(T a, T b, float t)` here as well.

chapters/slang.adoc
```diff
@@ -0,0 +1,744 @@
...
+
+* *Cross-API compatibility*: Write shaders that can be used with both Vulkan and DirectX with minimal changes
+* *Familiar syntax*: Developers with HLSL experience can leverage their existing knowledge
+* *Advanced language features*: Access to modern programming constructs like generics and interfaces
+* *Industry adoption*: Growing adoption in game engines and graphics applications
+* *Tooling support*: Rich ecosystem of tools and IDE integrations
+
+== Differences Between HLSL and Slang
+
+While Slang is built on HLSL, it adds several powerful features that make shader development more efficient, maintainable, and flexible.
```
juliusikkala: It sound a bit odd to say that "Slang is built on HLSL". To me, it feels like Slang would itself use or be lowered to HLSL (which it can do, but that's not what you'd do with Vulkan, where Slang is directly compiled to SPIR-V). Perhaps this could instead say "is largely compatible with HLSL 2018"?

chapters/slang.adoc
```diff
@@ -0,0 +1,744 @@
...
+
+=== Generics and Templates
+
+Slang adds full support for generics, similar to C# or Java, allowing for type-safe parameterized code:
+
+[source,slang]
+----
+// Generic function in Slang
+template<T>
+T min(T a, T b) {
```
juliusikkala: Should probably be `T min<T>(T a, T b)`, I don't think the `template<T>` syntax compiles.

chapters/slang.adoc
```diff
@@ -0,0 +1,744 @@
...
+// Compile-time reflection
+struct Material {
+    float4 baseColor;
+    float roughness;
+    float metallic;
+};
+// Get all fields of a type at compile time
+__generic<T>
+void bindMaterial(ParameterBlock<T> block, Material material) {
+    __for(field in getFields(T)) {
```
juliusikkala: What is this? I've never seen this in Slang code, and there's not a single `__for` in the entire compiler codebase. Almost like dreamed up by an LLM?

chapters/slang.adoc
```diff
@@ -0,0 +1,744 @@
...
+* *Cross-compilation*: Generate code for different shader stages from a single source
+* *Built-in reflection*: Generate reflection data during compilation
+* *Shader linking*: Link multiple shader modules together
+* *Diagnostic quality*: More detailed error messages and warnings
+
+Example of multi-target compilation:
+
+[source,bash]
+----
+slangc -profile glsl_spirv -entry main -stage vertex shader.slang -o shader.vert.spv
```
juliusikkala: I'm not sure if this is valid, highlighting to see if some other reviewer has opinions on this. I've used
`slangc -target spirv shader.slang -o shader.vert.spv` instead. Entry points can be marked in the source with an attribute like `[shader("compute")]` in front of the function. The `-entry main -stage vertex` is still valid though.

chapters/slang.adoc
```diff
@@ -0,0 +1,744 @@
...
+slangc -profile dxbc -entry main -stage vertex shader.slang -o shader.vert.dxbc
+----
+
+=== Runtime Behavior Differences
+
+Slang introduces some runtime behavior differences:
+
+* *Interface dispatch*: Runtime polymorphism through interfaces
```
juliusikkala: I think the more typical term for this is "Dynamic dispatch".

chapters/slang.adoc
```diff
@@ -0,0 +1,744 @@
...
+=== Step 6: Implement Advanced Metaprogramming
+
+Use Slang's metaprogramming capabilities for more powerful shader generation:
+
+[source,slang]
+----
+// Define shader permutations using compile-time parameters
+[shader("vertex")]
+[CompileTimeConstant(name="USE_NORMAL_MAPPING", type="bool")]
+[CompileTimeConstant(name="LIGHT_COUNT", type="int")]
```
juliusikkala: If these `CompileTimeConstant` attributes were a thing at some point, they sure aren't anymore.
I'm pretty sure this example could be written like this instead:
`VSOutput vertexShader<bool UseNormalMapping, int LightCount>(VSInput input) {`.

chapters/slang.adoc
```diff
@@ -0,0 +1,744 @@
...
+Migrating from GLSL to Slang involves more significant changes than migrating from HLSL, as the languages have different syntax and programming models. However, Slang's powerful features can make the transition worthwhile for many projects.
+
+[TIP]
+====
+For a comprehensive guide on migrating from GLSL to Slang, see the official link:https://docs.shader-slang.org/en/latest/coming-from-glsl.html[Coming from GLSL] documentation.
+====
+
+=== Key Differences to Consider
+
+* *Syntax*: GLSL uses a more procedural style similar to C, while Slang uses an object-oriented style similar to C++
```
juliusikkala: Hmm... Not sure if this is a good first impression to give, considering Slang has deprecated some OOP features like struct inheritance.

chapters/slang.adoc
```diff
@@ -0,0 +1,744 @@
...
+----
+
+Key command-line options include:
+
+* `-profile`: Specifies the target profile (use `glsl_spirv` for Vulkan)
+* `-entry`: Specifies the entry point function
+* `-o`: Specifies the output file
+* `-language`: Specifies the input language (default is Slang)
+* `-allow-glsl`: Enables GLSL as an input language
+* `-target`: Specifies the target platform (e.g., `vulkan`)
```
juliusikkala: There's some sort of a mix-up with `-target` and `-profile` going on here. Typically, `-target` specifies that we want SPIR-V (`-target spirv`) and `-profile` sets the SPIR-V version, e.g. `-profile spirv_1_4`.

chapters/slang.adoc
```diff
@@ -0,0 +1,744 @@
...
+Example of a well-organized Slang shader:
+
+[source,slang]
+----
+module Rendering.PBR;
+
+import Rendering.Common;
+import Rendering.Lighting;
+
+export interface IMaterial {
```
juliusikkala: `export`->`public` in this example as well.

chapters/slang.adoc
```diff
@@ -0,0 +1,744 @@
...
+    float metallic;
+
+    float3 evaluateBRDF(float3 viewDir, float3 lightDir, float3 normal) {
+        // PBR implementation
+        return calculatePBR(albedo, roughness, metallic, viewDir, lightDir, normal);
+    }
+}
+
+[shader("fragment")]
+export float4 fragmentMain(VSOutput input) {
```
juliusikkala: This should probably be in a different module. It could perhaps have a comment above it to mark that. At least to me, it doesn't seem well-organized to have something as re-usable as `IMaterial` in the same module with some specific fragment shader.

chapters/slang.adoc
```diff
@@ -0,0 +1,744 @@
...
+    // Shader implementation
+}
+----
+
+=== Performance Considerations
+
+* *Use generics judiciously*: While powerful, excessive use of generics can increase compile times
```
juliusikkala: Do we have actual numbers on this? Slang's generics are checked before specialization, so the compilation time characteristics should at least be a bit better than with C++ templates, where this statement would definitely be true.

chapters/slang.adoc
```diff
@@ -0,0 +1,744 @@
...
+* *Profile interface dispatch*: Test performance impact of interface-based polymorphism
```
juliusikkala: Another place where mentioning the flag that will hopefully get added in https://github.com/shader-slang/slang/pull/9273 would be useful.

chapters/slang.adoc
```diff
@@ -0,0 +1,744 @@
...
+* *Generics for reusable algorithms*: Write generic algorithms that work with different types
+
+Example of shader composition:
+
+[source,slang]
+----
+// In lighting.slang
+module Lighting;
+
+export interface ILight {
```
juliusikkala: `export` -> `public`

chapters/slang.adoc
```diff
@@ -0,0 +1,744 @@
...
+float4 calculateLighting(float3 position, float3 normal) {
+    // Implementation that works across APIs
+}
+----
+
+=== Shader Composition and Reuse
+
+Slang's module system enables powerful composition patterns:
+
+* *Interface-based composition*: Define interfaces for pluggable components
```
juliusikkala: It's odd to lead with
> Slang's module system enables powerful composition patterns:

and then start talking about interfaces and generics, when those can be used without modules as well. They're independent features.

juliusikkala (changes requested):
Hmm... I think this requires quite a bit of work still. There are lots of references to deprecated and sometimes even non-existent features. The text is also very repetitive at places. Many sections are simply lists with little substance. Lots of repetition of modules-interfaces-generics.

I hope other reviewers can fill in more details, especially about the overall structure. I focused more on correctness in code examples, where I likely also didn't find every problem, so I hope others will check those carefully as well.

chapters/slang.adoc
```diff
@@ -0,0 +1,744 @@
...
+
+SlangCompileRequest* compileRequest = nullptr;
+session->createCompileRequest(&compileRequest);
+
+// Add source code
+int translationUnitIndex = spAddTranslationUnit(compileRequest, SLANG_SOURCE_LANGUAGE_SLANG, "shaderCode");
+spAddTranslationUnitSourceString(compileRequest, translationUnitIndex, "shaderCode", shaderSource.c_str());
+
+// Set entry point
+spAddEntryPoint(compileRequest, translationUnitIndex, "main", SLANG_STAGE_FRAGMENT);
```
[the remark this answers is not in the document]
gpx1000: I gave it the best try at the best docs I could find several months ago to come up with this, so it's my best effort.  If no one able to chime in knows, I'm going to remove it to resolve this.  I don't want the wrong information in the guide.

chapters/slang.adoc
SaschaWillems: Exactly. That command line looks weird. Why use `glsl_spirv` when targeting Vulkan? The SPIR-V targets like `spirv_1_4` make more sense imo.

chapters/slang.adoc
```diff
@@ -0,0 +1,744 @@
...
+
+Example of cross-API shader:
+
+[source,slang]
+----
+// Common shader code
+module Shaders.PBR;
+
+// API-specific resource binding
+#if SLANG_VULKAN
```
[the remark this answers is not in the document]
SaschaWillems: Indeed, `SLANG_VULKAN` does not seem to exists.

Also shouldn't this be `#ifdef` or `#if defined`?

chapters/slang.adoc
```diff
@@ -0,0 +1,744 @@
...
+
+* *Use generics judiciously*: While powerful, excessive use of generics can increase compile times
+* *Profile interface dispatch*: Test performance impact of interface-based polymorphism
+* *Leverage compile-time evaluation*: Use Slang's metaprogramming to move work to compile time
+* *Consider specialization*: Generate specialized shader variants for performance-critical paths
+
+=== Vulkan-Specific Best Practices
+
+* *Use parameter blocks*: Organize resources into parameter blocks for better organization
+* *Explicit bindings*: Always specify explicit descriptor set and binding indices
```
SaschaWillems: I wouldn't consider that best practices. Not having to specify these is actually one of the best things about using Slang with Vulkan.

SaschaWillems:
Out of curiosity: Were some parts written by/with an AI?

Esp. the slang.adoc has lots of bullet lists with lots of repetitions, that make it read like it was at least partially written by an AI. E.g. the "multi api/cross compile" thing is mentioned at least 5 times in different places.

PR #334, opened 2025-07-22
Add in Windowing, Audio and Input chapter.

gpx1000:


spencer-lunarg:
The audio section starts to feel like the Vulkan Guide is not the right spot for it, while it is common for graphical application to want to use audio, I wouldn't expect to find "how use Vulkan" guide in a OpenAL article (nor would I find it as creditable as its not the domain you expect to find things)

gpx1000:
We can remove the audio section, but an upcoming tutorial on building a simple game engine goes into it including running the processing for it in the compute shader for HRTF.  The guide being the encyclopedia for that experience this is kinda a great place for it.  However, you're not wrong, the Guide itself as a wholly separate product does feel outside the wheelhouse of audio unless it's in the context of using Vulkan compute to do advanced audio things; which this doesn't cover.  We can review in meetings to discuss better placement if you prefer.

SaschaWillems:
Would also love to see [SFML](https://www.sfml-dev.org/index.php) in there. That's the one I've been using the most.

gpx1000:
I've never used SFML, would you mind providing some relevant details?  I can try to spend a few hours reviewing the tutorials and documentation to see if I can create enough to provide the requisite information.  Sadly, I write best when i write about things I know something about so it'll be hard for me to add that without more time gaining the requisite information.

SaschaWillems:
Sure, will do :)

SaschaWillems (dismissed):
As per today's call: LGTM.

Will add SFML once this is merged.

spencer-lunarg (dismissed):
will need a rebase, then lets merge

PR #338, opened 2025-08-04
Initial TBR chapter.

gpx1000:
NB, fix the TBR link to the Simple Game Engine tutorial when it is published.


chapters/tile_based_rendering_best_practices.adoc
```diff
@@ -0,0 +1,2202 @@
...
+Modern TBR implementations share common characteristics that affect optimization strategies:
+
+- **Tile Sizes**: Typically 16x16 or 32x32 pixels, with some supporting variable sizes
+- **Tile Memory**: Limited on-chip memory (256KB-1024KB) for storing tile data
+- **MSAA Efficiency**: TBR handles multisampling more efficiently due to tile memory resolve
+- **Attachment Optimization**: Depth attachments can often stay in tile memory (`VK_ATTACHMENT_STORE_OP_DONT_CARE`)
+- **Render Target Limits**: Optimal performance with 4-8 render targets depending on memory constraints
+
+**Adaptive Rendering Architecture:**
+Modern TBR implementations can dynamically switch between rendering modes based on workload characteristics:
```
zehui-l, line 35: As far as I know, there is only one TBR implementation that can switch between those modes, which is Qualcomm, to accommodate Windows requirement. If that is the case, it would not be appropriate to declare it as a modern TBR implementation.

chapters/tile_based_rendering_best_practices.adoc
```diff
@@ -0,0 +1,2202 @@
...
+- **Rendering Modes**: TBR (complex scenes), IMR (simple scenes), or Hybrid (mixed workloads)
+- **Decision Factors**: Vertex count thresholds (~100K vertices), draw call density, and scene complexity
+- **Configuration Strategy**: Optimize render passes based on chosen mode - maximize tile memory for TBR, minimize state changes for IMR
+- **Performance Benefits**: Automatic adaptation to workload characteristics improves overall efficiency
+
+**Advanced TBR Architecture Patterns:**
+Modern TBR implementations use sophisticated optimization strategies:
+
+- **Variable Tile Sizes**: Range from 16x16 to 64x64 pixels, with 32x32 being common default
+- **Tile Memory Management**: Typical range 256KB-1024KB, requires careful resource allocation
```
zehui-l, line 46: The resource allocation is done by Vulkan implementations, and is opaque to app developers.

chapters/tile_based_rendering_best_practices.adoc
```diff
@@ -0,0 +1,2202 @@
...
+- **Memory Type Selection**: TBR GPUs benefit from coherent memory (`VK_MEMORY_PROPERTY_HOST_COHERENT_BIT`) for tile data
```
zehui-l, line 47: Using coherent memory implies that there must be memory synchronization after every store operation or before every load operation, which is inefficient. I can't see its connection to tile data either, as tile data allocation is opaque to developers.

chapters/tile_based_rendering_best_practices.adoc
```diff
@@ -0,0 +1,2202 @@
...
+- **Subpass Optimization**: Optimal performance with 3 subpasses, maximum 8 render targets efficiently supported
```
zehui-l, line 48: Optimal number of subpasses and render targets varies depending on implementation, render target pixel format and other factors.

chapters/tile_based_rendering_best_practices.adoc
```diff
@@ -0,0 +1,2202 @@
...
+- **Render Pass Design**: Multiple subpasses with proper dependencies (`VK_DEPENDENCY_BY_REGION_BIT`) maximize tile memory utilization
+
+[[tbr-optimization-considerations]]
+=== TBR Optimization Considerations
+
+Modern TBR architectures share common optimization principles that can be applied generically across different hardware implementations:
+
+**Core TBR Optimization Principles:**
+
+- **Tile Memory Management**: TBR GPUs have limited tile memory that must be carefully managed
```
zehui-l, line 58: The tile memory is managed by Vulkan implementations, and is opaque to app developers.

cforfang:
Can I ask if a lot of this was written by AI? I'm very surprised by a lot of the text. Also the notion that one might as a SW developer 'choose' between TBR vs IMR, and have code trying to determine what to pick (ref the `PowerConsumptionAnalyzer` code) is very strange to me.

gpx1000:
Parts were written by AI, specifically the power consumption analyzer code as that is outside my normal wheelhouse; but looking at the references they looked solid and I edited it to make it read mostly correct to me.  So, genesis by AI sure, but heavy human editing.

cforfang:
Probably not too useful to drop more 'random' drive-by comments like this, but for another example I think none of the use-cases mentioned for VK_EXT_shader_tile_image (bloom, edge-detection, FXAA, SSR) makes sense as the extension only gives access to the _current_ pixel while all of these effects need access to _other_ pixels.

FWIW I've ping some folks here at Arm to see if we can help review and support development of the guide -- I think it's a great initiative to be clear, but it probably needs some close review especially as there is not too much good and up-to-date public info about current mobile GPUs (hence also why the idea of the guide is good, of course) :)

gpx1000:
Thanks it's MUCH appreciated.  I'm by far not the best expert at TBR; and I really want to try to get updated information out there.  There's a reason I read all of the research articles linked and tried to put as much research into this chapter as I could.  If we could get more details and more review, I'm much happier.  Soon as I get a chance, I'm going to update from the comments already generated here.

solidpixel:
The chapter title is Tile-Based Rendering Best Practices, but most of what it talks about is nothing to do with tile-based rendering but related to other aspects of vendor-specific detail or orthogonal mobile GPU issues (constant registers, coherent memory, thermal, etc). For a Vulkan guide I'd probably split this up - having a topic focused only on the effects of being tile based is useful and the rest is somewhat a distraction. 

The most important things for tilers (good use of loadOp/storeOp) seems to be buried right at the end, and the second most important (good use of pipeline barriers to get pipelining) isn't mentioned at all.

SaschaWillems:
Not that much of a hardware guy, but isn't laziliy allocated memory / transient attachment and important Vulkan concept for TBRs? If so might be good to add that.

SaschaWillems:
And I second the remarks about the power consumption part of that chapter. I tried to understand the code and data, but felt kinda lost. Wouldn't stuff like that require querying vendor specific apis to get real world power usage? Didn't see that mentioned anywhere.

SaschaWillems:
And other links don't make sense, e.g. this:

Vulkan-Hpp: [Modern C++ bindings with TBR optimization examples](https://github.com/KhronosGroup/Vulkan-Hpp)

That links to the Vulkan-Hpp headers, I don't see why or how that relates to TBR optimizations?

gpx1000:
I'm going to rewrite this.  Sorry not ready for prime time.

chapters/tile_based_rendering_best_practices.adoc
```diff
@@ -0,0 +1,2202 @@
...
+
+Several Vulkan extensions provide specific optimizations and capabilities for TBR architectures. This section provides concrete recommendations about what applications may benefit from these extensions:
+
+[[vk-ext-robustness2]]
+=== VK_EXT_robustness2
+
+This extension provides improved robustness when dangerous undefined behavior occurs, such as out-of-bounds array access. This is particularly important for TBR architectures where tile memory constraints can make buffer overruns more problematic.
+
+**Mobile developer guidance:**
+Mobile developers are strongly encouraged to use VK_EXT_robustness2 when targeting TBR GPUs, as tile memory constraints make out-of-bounds access more likely to cause visible artifacts or crashes.
```
iagoCL: No.

Mobile developers targetting Mali GPUs are encouraged not to use bounds checking.

Robust buffer access is a debugging feature and we recommend be temporarily enabled it to investigate application crashes or visual artifacts. Enabling it in production will negatively impact performance.

> Enabling bounds checking causes loss in performance for accesses to uniform buffers and shader storage buffers.

https://developer.arm.com/documentation/101897/0304/Buffers-and-textures/Robust-buffer-access?lang=en

zehui-l:
Huawei Maleoon GPU Guide: [Maleoon GPU Rendering Optimization](https://developer.huawei.com/consumer/en/doc/best-practices/bpta-maleoon-gpu-best-practices)

chapters/tile_based_rendering_best_practices.adoc
```diff
@@ -0,0 +1,2202 @@
...
+- **Tile Memory Optimization**: Efficient use of fast on-chip tile memory
+- **Adaptive Rendering**: Switch between rendering modes based on workload
+- **Geometry Binning**: Efficient spatial sorting of primitives
+
+**Tile Memory Management Strategies:**
+
+- **Memory Calculation**: Typical tile memory 512KB, calculate usage based on tile size (32x32 pixels), format size, and sample count
+- **Capacity Planning**: Ensure color + depth buffers fit within tile memory constraints
+- **Render Pass Optimization**: Use `VK_ATTACHMENT_STORE_OP_DONT_CARE` for depth when keeping in tile memory
+- **Fallback Strategy**: Switch to system memory rendering when tile memory insufficient
```
awalters-vk-img, line 80: Not controllable from an application using the Vulkan API

chapters/tile_based_rendering_best_practices.adoc
```diff
@@ -0,0 +1,2202 @@
...
+|https://developer.nvidia.com/rtx/[20 GB/s (typical desktop)]
+|https://www.imaginationtech.com/[**8x reduction with TBR**]
+
+|Power Consumption
+|https://developer.arm.com/documentation/102179/latest/[1.2W (mobile gaming)]
+|https://www.nvidia.com/en-us/geforce/graphics-cards/[2.4W (equivalent workload)]
+|https://developer.qualcomm.com/software/adreno-gpu-sdk/[**50% reduction with TBR**]
+
+|Overdraw Performance Impact
+|https://www.imaginationtech.com/[Minimal (resolved in tile memory)]
```
awalters-vk-img, line 170: Link is just going to the homepage, was there a particular article this was supposed to reference?

chapters/tile_based_rendering_best_practices.adoc
```diff
@@ -0,0 +1,2202 @@
...
+    }
+
+private:
+    static constexpr uint32_t FRAME_REUSE_THRESHOLD = 2; // Reuse after 2 frames
+    std::vector<TimeSlicedResource> timeSlicedResources_;
+};
+----
+
+[[caching-optimization]]
+=== Memory Caching Optimization
```
awalters-vk-img, line 1222: This is more of a UMA (unified memory architecture) vs. LMA (local memory architecture) difference rather than a TBR/IMR difference - not sure if that is worth separating into a chapter specific to that distinction instead.

chapters/tile_based_rendering_best_practices.adoc
```diff
@@ -0,0 +1,2202 @@
...
+
+    gl_FragColor = vec4(finalColor, 1.0);
+}
+----
+
+[[precision-optimization]]
+=== Half-Precision Float Optimization
+
+Using half-precision floats in shaders can speed up execution and reduce bandwidth on mobile TBR devices. Use low-precision numbers in fragment and compute shaders when visual quality permits:
+
```
awalters-vk-img, line 1684: Might be worth pointing out that mediump should be checked on as many devices as possible as they act as something of a hint - using mediump and testing only on one device that mayu under the hood still be using F32 can be very misleading and lead to visual issues on devices actually employing mediump.

I'd still recommend using it whenever possible, but it might be a worth while note/pointer

chapters/tile_based_rendering_best_practices.adoc
```diff
@@ -0,0 +1,2202 @@
...
+- **Tile Memory Management**: TBR GPUs have limited tile memory that must be carefully managed
+- **Early Fragment Rejection**: TBR architectures support efficient early fragment culling
+- **Bandwidth Optimization**: Critical for mobile and power-constrained devices
+
+**Bandwidth Optimization Strategies:**
+
+- **Attachment Configuration**: Final attachments use `VK_ATTACHMENT_STORE_OP_STORE`, intermediate attachments use `VK_ATTACHMENT_STORE_OP_DONT_CARE`
+- **Load Operations**: Use `VK_ATTACHMENT_LOAD_OP_CLEAR` for new content, `VK_ATTACHMENT_LOAD_OP_DONT_CARE` for intermediate results
+- **MSAA Efficiency**: TBR handles 4x MSAA efficiently due to tile memory resolve capabilities
```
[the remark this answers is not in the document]
gpx1000: Removed that advice to ensure it's clear.

chapters/tile_based_rendering_best_practices.adoc
gpx1000: Removed it to ensure the information I have in here is correct and something I've been able to verify.  If you recommend adding it back, I'd be happy to.

chapters/tile_based_rendering_best_practices.adoc
```diff
@@ -0,0 +1,2202 @@
...
+
+[[vulkan-extensions-comprehensive-guide]]
+== Vulkan Extensions Comprehensive Guide
+
+Several Vulkan extensions provide specific optimizations and capabilities for TBR architectures. This section provides concrete recommendations about what applications may benefit from these extensions:
+
+[[vk-ext-robustness2]]
+=== VK_EXT_robustness2
+
+This extension provides improved robustness when dangerous undefined behavior occurs, such as out-of-bounds array access. This is particularly important for TBR architectures where tile memory constraints can make buffer overruns more problematic.
```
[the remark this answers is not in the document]
gpx1000: Removed this section entirely to ensure we ONLY have TBR/IMR details.

chapters/tile_based_rendering_best_practices.adoc
```diff
@@ -0,0 +1,306 @@
...
+[[imr-architecture-analysis]]
+=== IMR Architecture Analysis
+
+Immediate Mode Rendering does not perform screen-space binning prior to rasterization. Fragment results are typically written to external memory as they are produced.
+
+Key characteristics (high-level):
+
+- No explicit on-chip tile memory model exposed to applications.
+- Overdraw tends to generate more external memory traffic than on tilers; minimizing overdraw is important.
+- Applications should rely on standard Vulkan techniques (early depth/stencil, appropriate load/store ops, and subpasses where helpful) and profile on target devices.
```
bacTlink: I am seeing "profile on target devices", "measure on target devices", "profiling results on target hardware" many times in this documentation. This kind of redundant phrases should be cleaned up.

bacTlink:
Currently information is scattered in various corners. And same information appears a few times including thing like "profile on target devices", or "Tile size not exposed by core Vulkan".
The documentation structure could be improved by establishing a main line of reasoning and developing the content within the framework of this logic.

chapters/tile_based_rendering_best_practices.adoc
```diff
@@ -0,0 +1,235 @@
...
+
+For a Vulkan developer, this architecture means that memory bandwidth is often the most significant performance factor.
+If you can keep your data within the tile memory, your application will be faster and consume less power.
+If you force the GPU to constantly "round-trip" data back and forth between the chip and external VRAM, performance will suffer.
+
+[[mobile-gpu-architectures]]
+== Understanding Tiler Architectures
+
+Mobile GPUs operate in power-constrained environments, which makes bandwidth efficiency critical.
+Since Vulkan hides many of the internal hardware details—like the exact size of a tile or how the GPU schedules work—the best way to optimize is to provide the driver with clear "intent."
```
solidpixel: Why quotes around "intent"? Doesn't seem to need them for this usage.

chapters/tile_based_rendering_best_practices.adoc
```diff
@@ -0,0 +1,224 @@
...
+You don't allocate "tile memory" directly.
+Instead, the number of attachments you use, their formats, and whether you enable MSAA all determine how much data needs to be stored per pixel.
+Because the total on-chip memory is fixed, using pixel formats with smaller bit depths can often allow the hardware to use larger tiles or avoid spilling data to external memory.
+If your attachments require more bits than the GPU has on-chip for a given tile size, the driver might have to use smaller tiles, which reduces efficiency.
+
+One often overlooked optimization for tilers involves the **binning pass**.
+Tilers usually process geometry twice: once to determine which triangles fall into which tiles (binning), and a second time to actually render the pixels.
+To speed up the binning pass, consider storing your vertex positions in a separate buffer from other attributes like UVs or normals.
+This allows the GPU to read only the data it needs to calculate tile coverage, significantly reducing unnecessary bandwidth.
+Ideally, position data should be stored as `highp` to ensure accuracy during this critical phase.
```
solidpixel: Disagree that input position needs to be "highp". 

The actual calculation in the shader definitely needs to be done as fp32, but vertex position data in memory only needs to be precise enough in memory to maintain sufficient accuracy in the model object space coordinates. Using unorm16 coordinates for a typical 10 meter real-world equivalent model gives 0.15 mm quantization accuracy which is plenty good enough for most use cases.

chapters/tile_based_rendering_best_practices.adoc
```diff
@@ -0,0 +1,224 @@
...
+== Optimization Strategies for TBR
+
+Most performance gains on a tiler come from answering one question: "Do I really need to move this data to external memory?"
+
+[[attachment-management]]
+=== Attachment Load and Store Ops
+
+Your primary tool for controlling bandwidth is the render pass attachment configuration.
+Whether you are using traditional `VkRenderPass` objects or the modern `VK_KHR_dynamic_rendering` extension, the principles are the same.
+The `loadOp` and `storeOp` settings are not just "cleanup" steps; they are direct instructions to the hardware.
```
solidpixel: why quotes around cleanup?

chapters/tile_based_rendering_best_practices.adoc
```diff
@@ -0,0 +1,224 @@
...
+
+If you are starting a new frame or a fresh pass, use `VK_ATTACHMENT_LOAD_OP_CLEAR`.
+This is significantly faster than using explicit clear commands like `vkCmdClearAttachments`, as it allows the hardware to initialize tile memory directly without any external memory traffic.
+If you know you are going to completely overwrite the tile's contents—for example, by rendering opaque geometry that covers the entire screen—you can use `VK_ATTACHMENT_LOAD_OP_DONT_CARE`.
+This tells the GPU it doesn't need to waste time loading the previous frame's data from memory OR performing a clear.
+
+Similarly, use `VK_ATTACHMENT_STORE_OP_DONT_CARE` for any attachment you don't need after the pass is finished.
```
solidpixel, line 65: or use `STORE_OP_NONE` if you want to indicate "don't write", but also don't want to logically discard the existing content in memory.

chapters/tile_based_rendering_best_practices.adoc
```diff
@@ -0,0 +1,224 @@
...
+Depth buffers and multisampled "resolve" sources are the most common candidates here.
```
solidpixel: Mentioning resolve here is likely going to make this more confusing to read because it immediately makes you think of resolve attachments, whereas this is about the other kind of attachment.  Just "depth attachments, stencil attachments, and multisampled color attachments"?

chapters/tile_based_rendering_best_practices.adoc
```diff
@@ -0,0 +1,224 @@
...
+By telling the GPU you don't care about the final state of these images, you prevent it from wasting bandwidth writing them back to main memory.
+
+When using **Dynamic Rendering** (`VK_KHR_dynamic_rendering`), you specify these same operations in the `VkRenderingInfo` structure.
+This extension simplifies your code by removing the need for render pass and framebuffer objects, but the hardware logic remains identical.
+You must remain disciplined about your load and store operations to avoid performance regressions.
+
+When using traditional render passes, try to structure them so that the driver can "merge" subpasses.
```
solidpixel, line 74: This is a bit of a throw away paragraph that really isn't explained anywhere. Subpass merging could be worth a bit more of a treatment than this, although obviously a bit of an evolutionary dead end.

chapters/tile_based_rendering_best_practices.adoc
```diff
@@ -0,0 +1,224 @@
...
+
+For intermediate data that only lives during a single render pass, you should mark your images with `VK_IMAGE_USAGE_TRANSIENT_ATTACHMENT_BIT`.
+When combined with memory that is `VK_MEMORY_PROPERTY_LAZILY_ALLOCATED_BIT`, some GPUs can avoid allocating any physical system memory for these attachments at all.
+The data exists only in the on-chip tile memory and is discarded when the pass ends.
+This is particularly useful for depth buffers or G-buffer attachments in a deferred renderer that are consumed in a later subpass.
+
+[[msaa-resolve-patterns]]
+=== Efficiency with MSAA
+
+Multisampling is particularly efficient on tilers because the "resolve" from 4x or 8x down to 1x can often happen entirely on-chip.
```
solidpixel: Or 2x or 16x. Why mention specific sizes?

chapters/tile_based_rendering_best_practices.adoc
```diff
@@ -0,0 +1,224 @@
...
+
+[[tbr-vs-imr-detailed-analysis]]
+== TBR vs IMR: A Practical Comparison
+
+It is helpful to contrast tilers with Immediate Mode Renderers (IMR), which are common in high-end desktop GPUs.
+
+* **IMR GPUs** typically process triangles and write the resulting fragments to memory almost immediately. They rely on high-bandwidth memory and large caches to handle the traffic. Overdraw on an IMR is expensive because every pixel written potentially triggers a memory write.
+* **TBR GPUs** defer those writes. By "binning" the geometry and processing by tile, they can perform many operations—like blending and depth testing—entirely within the tile memory. The memory write only happens once the tile is finished.
+
+While you shouldn't try to build a renderer that switches between TBR and IMR logic at runtime, understanding the difference helps you write code that is efficient for both. Good attachment management and avoiding unnecessary overdraw benefit every architecture, but they are absolutely essential for performance on a tiler.
```
solidpixel: Avoiding fragment overdraw is orthogonal to being a tiler (and if anything tilers are less susceptible to it costing performance because of better HSR).

chapters/tile_based_rendering_best_practices.adoc
```diff
@@ -0,0 +1,224 @@
...
+Incorrect use of pipeline barriers can break this overlap.
+If you use a barrier that is too broad—like `VK_PIPELINE_STAGE_ALL_COMMANDS_BIT`—you might force the GPU to finish all pending fragment work before it can even start the binning pass for the next set of draws.
+Instead, use the most specific stages and access masks possible.
+For example, if a compute shader produces data for a vertex buffer, the barrier should only synchronize the compute stage with the vertex input stage.
+This allows the tiler's geometry engine to start working as soon as the data is ready, even if the fragment units are still busy with a previous task.
+
+[[overdraw-and-sorting]]
+=== Managing Overdraw and Depth Logic
+
+Overdraw is one of the biggest bottlenecks on mobile GPUs.
```
solidpixel: Overdraw cost has nothing to do with being tile based. 

It's a problem on less performant mobile GPUs because the GPU is less powerful, so lots of blended overdraw means you can more easily become fragment bound, but that's nothing to do with the GPU being tile based. Delete?

chapters/tile_based_rendering_best_practices.adoc
```diff
@@ -0,0 +1,224 @@
...
+Even though writes are deferred, executing a fragment shader multiple times for the same pixel consumes valuable execution unit (EU) cycles and power.
+Sorting your opaque objects front-to-back is the most effective way to combat this.
```
solidpixel: As above, nothing to do with being tile based, so feels out of place.

chapters/tile_based_rendering_best_practices.adoc
```diff
@@ -0,0 +1,224 @@
...
+It allows the GPU's early depth testing to discard hidden fragments before the fragment shader even runs.
+To make this hardware optimization work effectively:
+
+*   Use a consistent `VkCompareOp` (usually `LESS` or `LESS_OR_EQUAL`) throughout your render pass. Frequent changes to the depth comparison function can disable hardware depth culling.
+*   Enable depth writing (`depthWriteEnable`) whenever possible. If it's disabled, the GPU cannot update its internal depth culling blocks, leading to more overdraw for subsequent draws.
+*   Avoid enabling "Alpha to Coverage" simultaneously with depth testing unless necessary, as it can consume extra hardware resources and reduce execution efficiency on some tilers.
+*   Avoid using `discard` or writing to `gl_FragDepth` in your shaders unless absolutely necessary, as these operations can force the GPU to disable "early" depth testing and wait for the fragment shader to finish before it can determine visibility.
+
+[[shader-concurrency]]
+=== Shader Complexity and Concurrency
```
solidpixel: Also nothing to do with being tile based. Delete?

chapters/tile_based_rendering_best_practices.adoc
```diff
@@ -0,0 +1,224 @@
...
+While this adds a small amount of overhead, the increase in EU concurrency can often lead to a net gain in frame rate.
+
+You can also help the compiler by keeping your shader logic organized.
+For example, grouping uniform-related calculations together and using vectorized memory access can often help the hardware execute instructions more efficiently.
+Avoiding deep branch nesting—ideally staying under eight levels—and using ternary operators or built-in functions like `clamp()` and `mix()` instead of complex `if-else` blocks will also help maintain high thread occupancy.
+
+[[precision-and-prefetch]]
+=== Precision and Texture Optimization
+
+Using `mediump` (16-bit) instead of `highp` (32-bit) in your shaders is a classic mobile optimization.
```
solidpixel, line 173: Also nothing to do with being tile based. Delete?

chapters/tile_based_rendering_best_practices.adoc
```diff
@@ -0,0 +1,224 @@
...
+Beyond the basics of attachment management, several other factors influence how well your application runs on a tiler.
+
+[[pipelining-and-barriers]]
+=== Pipelining and Barriers
+
+On a tiler, the GPU usually processes the "binning" pass for the entire render pass before it starts the "fragment" pass for the first tile.
+To keep the hardware busy, you want these two stages to overlap as much as possible—the GPU should be binning the next frame while it is still shading the current one.
+
+Incorrect use of pipeline barriers can break this overlap.
+If you use a barrier that is too broad—like `VK_PIPELINE_STAGE_ALL_COMMANDS_BIT`—you might force the GPU to finish all pending fragment work before it can even start the binning pass for the next set of draws.
```
solidpixel: Not sure how this renders - is this missing spaces around the "—" in "broad—like", etc (occurs in other places in the doc too),

solidpixel (changes requested):

chapters/tile_based_rendering_best_practices.adoc
iagoCL: **Nitpick:**  
I am not sure what *build a renderer that switches between TBR and IMR logic at runtime* means.

Note that some GPUs can change between IMR and TBR.

I would change the sentence to something like:  
*It is not necessary to write a custom path for TBR and IMR GPUs. In general, understanding how a tiler works will help you write code that is efficient for both architectures.*

chapters/tile_based_rendering_best_practices.adoc
iagoCL: *Sorting your opaque objects front-to-back is the most effective way to combat this.*

This was recommended on mobile for longer than on PC. While it is relevant for older GPUs, it is no longer always recommended for newer architectures.

See: https://developer.arm.com/community/arm-community-blogs/b/mobile-graphics-and-gaming-blog/posts/immortalis-g925-the-fragment-prepass

chapters/tile_based_rendering_best_practices.adoc
iagoCL: I would personally keep this section with some changes, assuming we are focusing on TBR and mobile GPUs. However, I have no problem deleting it if we want to focus exclusively on TBR optimizations.

Complex shaders are more likely to cause register spilling on mobile GPUs, but this is somewhat orthogonal to tile-based rendering.

Note that this recommendation also applies to desktop GPUs.

chapters/tile_based_rendering_best_practices.adoc
iagoCL: Similar, I would keep it with some changes.

I think this is orthogonal to tilers, but it can still benefit bandwidth, so using `mediump` is recommended on mobile GPUs.

I personally prefer explicit types such as `float16`.

Note that some desktop GPUs can also benefit from `float16`.

chapters/tile_based_rendering_best_practices.adoc
```diff
@@ -0,0 +1,224 @@
...
+
+You can also leverage **texture reuse patterns**.
+Many mobile GPUs can optimize scenarios where you sample multiple textures using the same coordinates, or sample the same texture with different offsets.
+Grouping these operations together—for example, by using a base coordinate and a set of constant offsets—allows the hardware to reduce memory requests and lower power consumption.
+Advanced samplers on some hardware can even perform operations like convolution, maximum, or minimum filters directly within the sampling unit if the shader is structured correctly.
+
+[[synchronization-and-subpasses]]
+=== Synchronization and Pipeline Flow
+
+Frequent synchronization points—like calling `vkQueueWaitIdle`—can cause the GPU to stall while waiting for the CPU, or vice versa.
```
iagoCL: Is this tiler specific? I think it a best practices for all GPUs?

chapters/tile_based_rendering_best_practices.adoc
```diff
@@ -0,0 +1,224 @@
...
+[[pipelining-and-barriers]]
+=== Pipelining and Barriers
+
+On a tiler, the GPU usually processes the "binning" pass for the entire render pass before it starts the "fragment" pass for the first tile.
+To keep the hardware busy, you want these two stages to overlap as much as possible—the GPU should be binning the next frame while it is still shading the current one.
+
+Incorrect use of pipeline barriers can break this overlap.
+If you use a barrier that is too broad—like `VK_PIPELINE_STAGE_ALL_COMMANDS_BIT`—you might force the GPU to finish all pending fragment work before it can even start the binning pass for the next set of draws.
+Instead, use the most specific stages and access masks possible.
+For example, if a compute shader produces data for a vertex buffer, the barrier should only synchronize the compute stage with the vertex input stage.
```
iagoCL, line 153: I would mention that `ALL_GRAPHICS`–`ALL_GRAPHICS` should generally be avoided. It’s better to separate `VERTEX_SHADER_BIT` and `FRAGMENT_SHADER_BIT`, and mark resources according to the stages that actually use them.

iagoCL:
A bit orthogonal to tilers, but it might be useful to explain that compute shaders can have lower effective bandwidth compared to the vertex and fragment stages on mobile GPUs.

Developers may want to move work to fragment and avoid enabling `USAGE_STORAGE` on images that do not need it, as this can disable compression: https://developer.arm.com/documentation/101897/0304/Buffers-and-textures/AFBC-textures-for-Vulkan

chapters/tile_based_rendering_best_practices.adoc
solidpixel: Mobile optimization is a whole different topic (and huge) - the Arm best practices guides > 100 pages because it's a hard problem. Keep this on topic, add other parallel topics on other best practices if enough vendors agree on them.

chapters/tile_based_rendering_best_practices.adoc
solidpixel: Keep it. As a different topic parallel to this one.

chapters/tile_based_rendering_best_practices.adoc
iagoCL: I agree it is better to keep them as separate topics, but not a strong opinion 

This version already has mobile optimizations orthogonal to TBR, so I am not sure if it is intended.

chapters/tile_based_rendering_best_practices.adoc
gpx1000: Thanks, adjusted it.

gpx1000:
https://github.com/KhronosGroup/Vulkan-Guide/pull/367  I added an embedded chapter here as it looked like some of the advice was starting to get away from TBR and into generic embedded programming.  That made me realize we had a gap.  Please help review that and see if we need to add things there or move some things from here into there or vice-versa.

chapters/tile_based_rendering_best_practices.adoc
```diff
@@ -0,0 +1,212 @@
...
+By telling the GPU you don't care about the final state of these images, you prevent it from wasting bandwidth writing them back to main memory.
+
+When using **Dynamic Rendering** (`VK_KHR_dynamic_rendering`), you specify these same operations in the `VkRenderingInfo` structure.
+This extension simplifies your code by removing the need for render pass and framebuffer objects, but the hardware logic remains identical.
+You must remain disciplined about your load and store operations to avoid performance regressions.
+
+When using traditional render passes, try to structure them so that the driver can "merge" subpasses.
+Subpass merging is a powerful feature where the driver combines multiple logical subpasses into a single hardware pass.
+This ensures that intermediate data (like G-buffer attributes in a deferred renderer) stays entirely within the tile memory and is never written to external VRAM.
+While modern extensions like `VK_KHR_dynamic_rendering_local_read` provide similar benefits for dynamic rendering, subpass merging remains the most reliable way to achieve peak efficiency in complex pipelines using traditional render passes.
```
iagoCL: Nitpick: I find this sentence a bit confusing, it seems to imply that VK_KHR_dynamic_rendering_local_read is worse for subpass merging / performance.

Maybe something like this:

> Subpasses can be key to achieving good performance with traditional rendering pipelines using render passes, but this is not guaranteed, as subpass merging is implementation-dependent. Applications can use VK_EXT_subpass_merge_feedback to check whether their subpasses are merged. Modern pipelines using dynamic rendering can obtain similar benefits to subpasses by using VK_KHR_dynamic_rendering_local_read.

chapters/tile_based_rendering_best_practices.adoc
```diff
@@ -0,0 +1,212 @@
...
+It is better to separate `VK_PIPELINE_STAGE_VERTEX_SHADER_BIT` and `VK_PIPELINE_STAGE_FRAGMENT_SHADER_BIT`, and mark resources according to the stages that actually use them.
+If you use a barrier that is too broad — like `VK_PIPELINE_STAGE_ALL_COMMANDS_BIT` — you might force the GPU to finish all pending fragment work before it can even start the binning pass for the next set of draws.
+Instead, use the most specific stages and access masks possible.
+For example, if a compute shader produces data for a vertex buffer, the barrier should only synchronize the compute stage with the vertex input stage.
+This allows the tiler's geometry engine to start working as soon as the data is ready, even if the fragment units are still busy with a previous task.
+
+[[depth-and-hsr]]
+=== Depth Testing and Hidden Surface Removal
+
+Tilers are designed to minimize the impact of overdraw through hardware-based Hidden Surface Removal (HSR) or forward pixel killing.
```
iagoCL: Nitpick:
- Designed is a strong word.
- Overdraw is not free on tilers and it still has a cost
- Not all TBRs GPUs have support / same support for HSR or forward pixel killing. Is important to profile on multiple hardware.

chapters/tile_based_rendering_best_practices.adoc
```diff
@@ -0,0 +1,212 @@
...
+[[precision-and-prefetch]]
+=== Precision and Texture Optimization
+
+Using `mediump` (16-bit) instead of `highp` (32-bit) in your shaders is a classic mobile optimization.
+While this is technically orthogonal to the tiling process, it is highly recommended for mobile GPUs because it significantly reduces memory bandwidth and register pressure.
+Many developers prefer using explicit types, such as `float16`, where supported by hardware and extensions.
+Note that some modern desktop GPUs also benefit from `float16` throughput and storage.
+
+Treat `mediump` as a hardware hint, not a guarantee.
+Some GPUs might still run at full 32-bit precision internally, while others will show visible artifacts.
```
iagoCL, line 179: I would add a small note to verify performance of your float16 optimizations.

Adding float16 to everything without careful consideration is usually not a good idea.

In some cases, f16 just introduces unnecessary conversions (f32 -> f16 -> f32). A compiler might try to detect this and keep everything in f32 if it is faster, most mobile GPUs are good with this optimization, but this is heuristic-based and can fail. I would recommend to profile F16 across multiple GPUs, tools like [Mali offline compiler](https://developer.arm.com/Tools%20and%20Software/Mali%20Offline%20Compiler) might be useful

PR #344, opened 2025-09-17
Add a Descriptor Buffer chapter

spencer-lunarg:
Adds a small chapter around `VK_EXT_descriptor_buffer` mainly from my  experiences trying to parse the many "buffer" and "binding" terms we have in the spec


chapters/descriptor_buffer.adoc
```diff
@@ -0,0 +1,250 @@
...
+
+VkDescriptorSetLayout ds_layout_0(bindings_a); // Set 0
+VkDescriptorSetLayout ds_layout_1(bindings_a); // Set 1
+VkDescriptorSetLayout ds_layout_2(bindings_b); // Set 2
+VkPipelineLayout pipeline_layot([ds_layout_0, ds_layout_1, ds_layout_2]);
+----
+
+== Query Descriptor Set Layout Sizes
+
+Now using the new `vkGetDescriptorSetLayoutSizeEXT` and `vkGetDescriptorSetLayoutBindingOffsetEXT` commands, we can get info from the driver what size it needs to properly use these `VkDescriptorSetLayout`.
```
gpx1000, line 81: I *generally* try to avoid words like "new" when referencing API calls.  That's a quick way to help ensure that the guide doesn't get dated as soon as those functions are no longer "new."  i.e. if someone were to look up those functions in the spec and see they were introduced in 1.4.x and the current spec is 1.5.x etc they might start to expect that the information is no longer potentially relevant.

chapters/descriptor_buffer.adoc
```diff
@@ -0,0 +1,250 @@
...
+
+[NOTE]
+====
+Don't make the assumption that `binding 0` will be the lowest offset!
```
gpx1000, line 85: I recommend giving a "why" when making statements like this.  What would cause a binding of 0 to not be the lowest, what would a good expected lowest number be?

chapters/descriptor_buffer.adoc
```diff
@@ -0,0 +1,250 @@
...
+
+== Creating Descriptor Buffers
+
+We will create a "special" `VkBuffer` with `VK_BUFFER_USAGE_RESOURCE_DESCRIPTOR_BUFFER_BIT_EXT` that now makes it a "descriptor buffer" `VkBuffer`.
+
+These buffers can be large and they hold a "look up table" to your resources and samplers.
+
+[NOTE]
+====
+The minimum required limit for `VkPhysicalDeviceDescriptorBufferPropertiesEXT::descriptorBufferAddressSpaceSize` is 128MB but link:https://vulkan.gpuinfo.org/displayextensionproperty.php?extensionname=VK_EXT_descriptor_buffer&extensionproperty=descriptorBufferAddressSpaceSize&platform=all[many devices can support 4GB]
```
gpx1000, line 112: If possible, try keeping them in the docs.vulkan.org site.  This will ensure that all browsers keep the site index in cache and keep the site fast/smooth.

gpx1000 (commented):
Great addition.  Very clear and clean explanation for descriptor_buffer.  Minor things that might be worthy changes are recommended.

chapters/descriptor_buffer.adoc
spencer-lunarg: I guess I meant "new" as in relative to what this extension adds, will re-word it

gpx1000 (approved):

PR #351, opened 2025-09-29
Update descriptor_buffer.adoc

spencer-lunarg:
Small extra gotcha around using `vkGetDescriptorSetLayoutSize`


spencer-lunarg:
update - CI is failing because `http://vktracer.com/` is offline, will give in a while before I deem it removed because it is a dead project

PR #368, opened 2026-03-16
Add profiling chapter with CPU/GPU optimization strategies and tooling

gpx1000:
Create a profiling guide for Vulkan.


SaschaWillems:
The guide already has some information on profiling in the [debugging tools chapter](https://github.com/KhronosGroup/Vulkan-Guide/blob/main/chapters/development_tools.adoc#profiling). Might be worth unifying that information to avoid duplications.

gpx1000:
Good call, I had forgotten about that.  I'll work on unifying.

spencer-lunarg:
I would be happy if we took `development_tools.adoc#profiling` and had it "point" this this chapter, like give a small bit of info and then go to the nice new dedicated chapter on it

gpx1000:
That sounds like a plan.  I'll just do that.  Need to finish one of the tutorial things am working on and will come back to this in the guide.

SaschaWillems (approved):

PR #384, opened 2026-06-12
Add VK_KHR_extended_flags chapter

spencer-lunarg:
@r-potter @oddhack if you could give this a review, the goal is just having "something" that explains why there is no `VkFormatFeatureFlags3`


oddhack (approved):
LGTM. I take it you are focusing on the '*4' issue and not doing complete API coverage of the extension. Maybe you would want to mention VkImageCreateFlagBits2KHR as well, since the new FormatFeature and ImageUsage flags are mentioned? But not a blocker.

spencer-lunarg:
merging, better to have than nothing, can always be improved if desired

